PR #401, opened 2014-12-10
Cleaner multi rom match autoselect

Zaneris:
Added a country code enum to the project that allows knowing what suffix
to look for in the name of the rom.


fzurita:
Hmm.... this should probably get fixed, it can lead to the wrong "Good name" for a ROM being displayed.

fzurita:
I went ahead and did a manual merge of this pull request: See here:

https://github.com/mupen64plus-ae/mupen64plus-ae/commit/e13c7e6bc2d6df3f02755e00d8ef24563328c895

Issue #405, opened 2015-02-14
Touchscreen button auto-hold not implemented?


BonzaiThePenguin:
I just discovered that Mupen64plus-AE *has* been updated since January 2014, but only if you compile it yourself! The new version is a lot better than the old version, thanks a lot for the work thus far.

One thing I can't seem to get working is the auto-hold feature. I duplicated a touchscreen layout and enabled the Auto-holdable option for the Z button, I set the touchscreen layout for a game to the layout I just created, then in the Global Settings I tried setting the button auto-hold to "On slide-off" *and* "On long-press", but neither seems to work as it should. No amount of sliding or holding on the Z button will get it to be recognized.

I don't remember the old version having a touchscreen layout editor at all, or an auto-hold option for that matter, so I guess this feature just hasn't been implemented yet?

BonzaiThePenguin:
Hm, nope, looks like it's implemented:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/c0ed117b2468e997325c83202b9fbe92cf14d5b0/src/paulscode/android/mupen64plusae/input/TouchController.java

    switch( mAutoHoldMethod )
    {
    case AUTOHOLD_METHOD_LONGPRESS:
        // Using long-press method, release auto-hold button
        if( mListener != null )
            mListener.onAutoHold( false, prevIndex );
        setTouchState( prevIndex, false );
        break;
        
    case AUTOHOLD_METHOD_SLIDEOUT:
        // Using slide-off method, engage auto-hold button
        if( mVibrator != null )
        {
            mVibrator.cancel();
            mVibrator.vibrate( AUTOHOLD_VIBRATE_PATTERN, -1 );
        }
        if( mListener != null )
            mListener.onAutoHold( true, prevIndex );
        setTouchState( prevIndex, true );
        break;
    }

I'll go try to figure out why this code isn't firing properly.

BonzaiThePenguin:
Aha, I found it!

Most of the app uses `touchscreenAutoHoldables` to serialize the autoholdable data, but when it comes time for GamePrefs to load the autoholdables, it uses `touchscreenAutoHoldables`. Note that the H is erroneously capitalized!

littleguy77:
Arg.  One of the things that really bugs me about the Android framework.  I wish they architected it in a way that allowed the compiler to catch it.

littleguy77:
Thanks for the PR!

PR #407, opened 2015-02-14
Fixed auto-hold for touchscreen controls

BonzaiThePenguin:
The auto-hold values were loaded using the wrong tag name when it came time to play the game, so auto-hold never worked. This just fixes the tag name.


BonzaiThePenguin:
I'm not sure how to reference an issue, but it's this one:

https://github.com/mupen64plus-ae/mupen64plus-ae/issues/405

littleguy77:
Thanks a lot!  I guess no one has used that feature in a while LOL.

Btw we have been publishing updates, just not on google Play.  We're still in the alpha stage for version 3.
http://www.paulscode.com/forum/index.php?topic=1818

BonzaiThePenguin:
Oh! I had searched for Mupen64Plus-AE *nightly*, but not *alpha*. Thanks for merging the PR!

PR #412, opened 2015-02-23
Adding localization support to the ROM scan feature

BonzaiThePenguin:
Updated CacheRomInfoTask to route through the localization system, then translated the strings to the best of my abilities. Also fixed a few errors with some of the files (copy-pasted French translations in non-French titles).

### DIFF
--- a/src/paulscode/android/mupen64plusae/task/CacheRomInfoTask.java
+++ b/src/paulscode/android/mupen64plusae/task/CacheRomInfoTask.java
@@ -127,7 +127,7 @@ public class CacheRomInfoTask extends AsyncTask<Void, ConfigSection, ConfigFile>
             mProgress.setMaxSubprogress( 0 );
             mProgress.setSubtext( "" );
             mProgress.setText( file.getAbsolutePath().substring( mSearchPath.getAbsolutePath().length() ) );
-            mProgress.setMessage( "Searching…" );
+            mProgress.setMessage( R.string.cacheRomInfo_searching );
             
             if( isCancelled() ) break;
             RomHeader header = new RomHeader( file );
@@ -147,7 +147,7 @@ public class CacheRomInfoTask extends AsyncTask<Void, ConfigSection, ConfigFile>
                     {
                         ZipEntry zipEntry = entries.nextElement();
                         mProgress.setSubtext( zipEntry.getName() );
-                        mProgress.setMessage( "Searching zip file…" );
+                        mProgress.setMessage( R.string.cacheRomInfo_searchingZip );
                         
                         if( isCancelled() ) break;
                         try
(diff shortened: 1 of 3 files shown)


BonzaiThePenguin:
Also for some reason the strings ended with the <?> glyph, so I changed that to …

littleguy77:
Thanks!

Did you use Google translate or something?  We use Transifex for translations.  All you need to provide are English translations, our translator teams take care of the rest.
https://www.transifex.com/projects/p/mupen64plus-ae/

src/paulscode/android/mupen64plusae/task/CacheRomInfoTask.java
```diff
@@ -103,6 +104,7 @@ public CacheRomInfoTask( Activity activity, File searchPath, String databasePath
     private final boolean mClearGallery;
     private final CacheRomInfoListener mListener;
     private final ProgressDialog mProgress;
+    private final Activity mActivity;
```
littleguy77: Rather than caching the activity and doing it this way, I'd prefer to just add another public method to ProgressDialog.java:

```
    public void setMessage( final int resid )
    {
        mActivity.runOnUiThread( new Runnable()
        {
            @Override
            public void run()
            {
                mTextMessage.setText( resid );
            }
        } );
    }
    ```

File: src/paulscode/android/mupen64plusae/task/CacheRomInfoTask.java [hunk 2]
```diff
@@ -127,7 +129,7 @@ protected ConfigFile doInBackground( Void... params )
             mProgress.setMaxSubprogress( 0 );
             mProgress.setSubtext( "" );
             mProgress.setText( file.getAbsolutePath().substring( mSearchPath.getAbsolutePath().length() ) );
-            mProgress.setMessage( "Searching�" );
+            mProgress.setMessage( mActivity.getString( R.string.cacheRomInfo_searching ) );
```
littleguy77: Then it's just
```
mProgress.setMessage( R.string.cacheRomInfo_searching );
```

BonzaiThePenguin:
Yeah, I used Google Translate. Attempted to make sure I got the connotations and conjugations correct, like making sure it didn't translate to downloading (bed) covers instead of cover art. Could only do so much, though! You're of course 100% welcome to delete it all and fill it in with the transifex stuff.

littleguy77:
Sorry you did all that work for nothing.  Would you mind rebasing this pull request and removing all the modified translation files?  Just keep the original (English) file and the modified java code.

src/paulscode/android/mupen64plusae/GalleryActivity.java
```diff
@@ -313,7 +326,14 @@ private void refreshGrid( ConfigFile config )
...
-                items.add( new GalleryItem( this, md5, goodName, romPath, artPath ) );
+                
+                String useName = goodName;
+                String baseName = config.get( md5, "baseName" );
+                if (baseName != null && baseName.length() != 0)
+                    useName = baseName;
+                if (romNames.containsKey(useName) && romNames.get(useName) == Boolean.TRUE)
+                    useName = goodName;
+                items.add( new GalleryItem( this, md5, useName, romPath, artPath ) );
             }
```
littleguy77: What is the purpose of this code?

Also, FYI the current main menu is just a stand-in.  Polishing will be done once @paulscode has built the new main menu.

BonzaiThePenguin:
I was working on a different feature unrelated to the translations. Didn't realize it'd show up in this pull request, sorry.

BonzaiThePenguin:
Alright, all fixed. Here's what that chunk of code did:

screenshot_2015-02-23-16-15-46

It didn't change the UI code at all, only the data that was passed into it. It just stripped the region codes from the ROM titles unless there were two ROMs for the same game.

But, I'll keep it out for now.

littleguy77:
Thanks much!

I recommend holding of on the ROM naming thing, since that is just a temporary UI and there are a lot of design decisions left to be made... best to do them wholesale rather than piecemeal.

If you're ever interested in contributing to the UI, we could definitely use more devs.  Take a look at the brainstorming thread below if you want to understand the current design choices and how they were reached:
http://www.paulscode.com/forum/index.php?topic=1130.0

PR #413, opened 2015-02-24
Fixed loading the auto hold images

BonzaiThePenguin:


BonzaiThePenguin:
Just as a heads up, right after this I have a PR that implements drag and drop support for the touchscreen controller editor.

src/paulscode/android/mupen64plusae/input/map/VisibleTouchMap.java
```diff
@@ -551,12 +551,12 @@ private void loadFpsIndicator( Profile profile )
...
-        Matcher matcher = Pattern.compile( "([^-]*)-hold(.*)" ).matcher( name );
-        if( matcher.groupCount() < 3 )
+        if ( !name.contains("-hold") )
             return;
         
-        String group = matcher.group( 1 );
-        String hold = matcher.group( 2 );
+        String[] fields = name.split( "-hold" );
+        String group = fields[0];
+        String hold = fields[1];
```
littleguy77: What's an example of a filename that this fixes?

BonzaiThePenguin:
I only used autohold on the Z button.

BonzaiThePenguin:
However none of them worked. Matcher always had a group count of 2 and returned before loading anything.

src/paulscode/android/mupen64plusae/input/map/VisibleTouchMap.java
BonzaiThePenguin: Ugh, I really need to learn my way around GitHub PRs – keep hitting the wrong reply button. None of the images loaded correctly since Matcher always returned a group count of 2, and Pattern.compile.matcher seemed like overkill for a simple field delimiter anyway so I just switched it out to split. All of the images load correctly now.

However, one example is buttonL-holdL.png, and another is groupAB-holdB.png.

littleguy77: Ah, thanks for tracking this down.  The bug is that I should have tested `!matcher.matches()` rather than `matcher.groupCount(0) < 3`.

I forget why I used a regex in the first place.  Maybe a carryover from when the parsing was more complex (the touchscreen stuff changed a lot between v2 and v3).  Your fix is easier to read than a regex, so I'll merge this.

Thanks :)

PR #415, opened 2015-02-24
Drag and drop editing for the touchscreen controls

BonzaiThePenguin:
Things that are changed with this commit:

- added drag and drop editing for the on-screen controller layout.
Right now the values are clamped to the nearest 5% to match the
behavior of the edit dialog, although I feel like we should allow any %
for that.
- tapping on a button plays the standard button sound effect, as I felt
that was missing.
- made it possible to swipe from the top or bottom of the screen to
show the back/home/menu buttons or status bar without *also* triggering
the edit dialog for a button that was under your finger, since that
dialog prevented the user from being able to use the menus at the top
of the screen.

### DIFF
--- a/src/paulscode/android/mupen64plusae/input/map/TouchMap.java
+++ b/src/paulscode/android/mupen64plusae/input/map/TouchMap.java
@@ -282,12 +282,8 @@ public class TouchMap
     {
         for( int i = 0; i < buttonNames.size(); i++ )
         {
-            String name = buttonNames.get( i );
-            if ( name.equals( assetName ) )
-            {
-                Image mask = buttonMasks.get( i );
-                return new Rect( mask.x, mask.y, mask.x + (int) ( mask.width * mask.scale ), mask.y + (int) ( mask.height * mask.scale ));
-            }
+            if ( buttonNames.get( i ).equals( assetName ) )
+                return new Rect( buttonMasks.get( i ).drawRect );
         }
         return new Rect(0, 0, 0, 0);
     }
@@ -360,10 +356,9 @@ public class TouchMap
      */
     public Rect getAnalogFrame()
     {
-        if( analogBackImage == null )
-            return new Rect(0, 0, 0, 0);
-        
-        return new Rect( analogBackImage.x, analogBackImage.y, analogBackImage.x + (int) ( analogBackImage.width * analogBackImage.scale ), analogBackImage.y + (int) ( analogBackImage.height * analogBackImage.scale ) );
+        if( analogBackImage != null )
+            return new Rect( analogBackImage.drawRect );
+        return new Rect(0, 0, 0, 0);
     }
     
     /**
--- a/src/paulscode/android/mupen64plusae/input/map/TouchMap.java
+++ b/src/paulscode/android/mupen64plusae/input/map/TouchMap.java
@@ -32,6 +32,7 @@ import paulscode.android.mupen64plusae.util.Image;
 import paulscode.android.mupen64plusae.util.Utility;
 import android.content.res.Resources;
 import android.graphics.Point;
+import android.graphics.Rect;
 import android.util.FloatMath;
 import android.util.Log;
 import android.util.SparseArray;
@@ -82,6 +83,9 @@ public class TouchMap
     
     /** Y-coordinates of the buttons, in percent. */
     private final ArrayList<Integer> buttonY;
+    
+    /** names of the buttons. */
+    private final ArrayList<String> buttonNames;
     
     /** Analog background image (fixed). */
     protected Image analogBackImage;
@@ -177,6 +181,7 @@ public class TouchMap
         buttonMasks = new ArrayList<Image>();
         buttonX = new ArrayList<Integer>();
         buttonY = new ArrayList<Integer>();
+        buttonNames = new ArrayList<String>();
     }
     
     /**
@@ -188,6 +193,7 @@ public class TouchMap
         buttonMasks.clear();
         buttonX.clear();
         buttonY.clear();
+        buttonNames.clear();
         analogBackImage = null;
         analogForeImage = null;
         analogBackX = analogBackY = 0;
@@ -265,6 +271,28 @@ public class TouchMap
     }
     
     /**
+     * Gets the frame for the N64 button with a given asset name
+     * 
+     * @param assetName The asset name for the button
+     * 
+     * @return The frame for the N64 button with the given asset name
+     * 
+     */
+    public Rect getButtonFrame( String assetName )
+    {
+        for( int i = 0; i < buttonNames.size(); i++ )
+        {
+            String name = buttonNames.get( i );
+            if ( name.equals( assetName ) )
+            {
+                Image mask = buttonMasks.get( i );
+                return new Rect( mask.x, mask.y, mask.x + (int) ( mask.width * mask.scale ), mask.y + (int) ( mask.height * mask.scale ));
+            }
+        }
+        return new Rect(0, 0, 0, 0);
+    }
+    
+    /**
      * Gets the N64 button mapped to a given mask color.
      * 
      * @param color The mask color.
@@ -323,6 +351,19 @@ public class TouchMap
         int dY = yLocation - ( analogBackImage.y + (int) ( analogBackImage.hHeight * scale ) );
         
         return new Point( dX, dY );
+    }
+    
+    /**
+     * Gets the N64 analog stick's frame.
+     * 
+     * @return The analog stick's frame.
+     */
+    public Rect getAnalogFrame()
+    {
+        if( analogBackImage == null )
+            return new Rect(0, 0, 0, 0);
+        
+        return new Rect( analogBackImage.x, analogBackImage.y, analogBackImage.x + (int) ( analogBackImage.width * analogBackImage.scale ), analogBackImage.y + (int) ( analogBackImage.height * analogBackImage.scale ) );
     }
     
     /**
@@ -506,6 +547,7 @@ public class TouchMap
             // Position (percentages of the digitizer dimensions)
             buttonX.add( x );
             buttonY.add( y );
+            buttonNames.add( name );
             
             // Load the displayed and mask images
             buttonImages.add( new Image( mResources, skinFolder + "/" + name + ".png" ) );
--- a/src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
+++ b/src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
@@ -413,7 +413,7 @@ public class TouchscreenProfileActivity extends Activity implements OnTouchListe
                 // unfortunately KitKat lacks a way to do this on its own,
                 // so just ignore all touches along the edges.
                 View view = getWindow().getDecorView();
-                if (y < 10 || y > view.getHeight() - 10 || x < 10 || x > view.getWidth() - 10)
+                if ( y < 10 || y > view.getHeight() - 10 || x < 10 || x > view.getWidth() - 10 )
                     return false;
             }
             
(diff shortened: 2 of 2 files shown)


src/paulscode/android/mupen64plusae/input/map/TouchMap.java
```diff
@@ -265,6 +271,28 @@ public int getButtonPress( int xLocation, int yLocation )
...
     
     /**
+     * Gets the frame for the N64 button with a given asset name
+     * 
+     * @param assetName The asset name for the button
+     * 
+     * @return The frame for the N64 button with the given asset name
+     * 
+     */
+    public Rect getButtonFrame( String assetName )
```
BonzaiThePenguin: In order to support drag and drop I needed a way to get the draw frames for the buttons and analog stick, and *that* needed a way to match the button layout index to the N64 index. Not sure if there's a better way?

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,120 @@ private void hideSystemBars()
...
+            initialY = y;
+            dragIndex = TouchMap.UNMAPPED;
+            dragging = false;
             
-            // D-pad buttons are not holdable
-            if( DPAD.equals( assetName ) )
-                index = -1;
+            // ignore touches that start at the very bottom or top of the screen
+            View view = getWindow().getDecorView();
+            if (y < 10 || y > view.getHeight() - 10)
```
BonzaiThePenguin: Can this be improved? I just used 10 pixels because it needed some sort of dead zone for swipes, without being so large that it interferes with the ability to tap on a button to edit it.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,120 @@ private void hideSystemBars()
...
+            else
+            {
+                // See if analog was pressed
+                Point point = mTouchscreenMap.getAnalogDisplacement( x, y );
+                int dX = point.x;
+                int dY = point.y;
+                float displacement = FloatMath.sqrt( ( dX * dX ) + ( dY * dY ) );
+                if( mTouchscreenMap.isInCaptureRange( displacement ) )
+                {
+                    dragIndex = -2;
```
BonzaiThePenguin: Should we use a constant for this? I just chose a number that wasn't being used, since the analog stick doesn't seem to have its own index value.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,120 @@ private void hideSystemBars()
...
+                    if (displacement >= 10)
+                        dragging = true;
+                }
+                if (!dragging)
+                    return false;
+                
+                // drag this button or analog stick around
+                
+                // calculate the X and Y percentage
+                View view = getWindow().getDecorView();
```
BonzaiThePenguin: Is this right? The values used in TouchMap.resize always seem to equal the size of the decorView, but maybe not? Not sure if I should be polling for these values from somewhere else.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
-            // D-pad buttons are not holdable
-            if( DPAD.equals( assetName ) )
-                index = -1;
+            if( AppData.IS_KITKAT && mUserPrefs.isImmersiveModeEnabled )
+            {
+                // ignore edge swipes.
+                // unfortunately KitKat lacks a way to do this on its own,
+                // so just ignore all touches along the edges.
+                View view = getWindow().getDecorView();
+                if (y < 10 || y > view.getHeight() - 10 || x < 10 || x > view.getWidth() - 10)
```
BonzaiThePenguin: http://stackoverflow.com/questions/20530333/ignore-immersive-mode-swipe

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
+                
+                // drag this button or analog stick around
+                
+                // calculate the X and Y percentage
+                View view = getWindow().getDecorView();
+                int newDragX = (x - (initialX - dragFrame.left)) * 100/(view.getWidth() - (dragFrame.right - dragFrame.left));
+                int newDragY = (y - (initialY - dragFrame.top)) * 100/(view.getHeight() - (dragFrame.bottom - dragFrame.top));
+                
+                // round to the nearest 5%
+                newDragX = Math.min(Math.max(Math.round(newDragX / 5) * 5, 0), 100);
```
BonzaiThePenguin: Really seems like this should be changed to 1%, especially on larger screens!

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
+        {
+            // if this touch was part of a drag/swipe gesture then don't tap the button
+            if (dragging)
+                return false;
+            
+            // show the editor for the tapped button
+            if (dragIndex == -2)
+            {
+                // play the standard button sound effect
+                View view = getWindow().getDecorView();
```
BonzaiThePenguin: From what I can gather online, this is automatically disabled by the OS if view sound effects are disabled globally. So it shouldn't need to check for any preferences here.

littleguy77:
Excellent, I'll take a look and answer your questions when I get home from work tonight.  At a glance it looks clean and well done, so I'll look forward to it.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
+                
+                // drag this button or analog stick around
+                
+                // calculate the X and Y percentage
+                View view = getWindow().getDecorView();
+                int newDragX = ( x - ( initialX - dragFrame.left ) ) * 100/( view.getWidth() - ( dragFrame.right - dragFrame.left ) );
+                int newDragY = ( y - ( initialY - dragFrame.top ) ) * 100/( view.getHeight() - ( dragFrame.bottom - dragFrame.top ) );
+                
+                // round to the nearest 5%
+                newDragX = Math.min( Math.max( Math.round( newDragX / 5 ) * 5, 0 ), 100 );
```
BonzaiThePenguin: Really seems like this should be changed to 1%, especially on larger screens!

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
+            // if this touch was part of a drag/swipe gesture then don't tap the button
+            if ( dragging )
+                return false;
+            
+            // show the editor for the tapped button
+            if ( ANALOG.equals( dragAsset ) )
+            {
+                // play the standard button sound effect
+                View view = getWindow().getDecorView();
+                view.playSoundEffect( SoundEffectConstants.CLICK );
```
BonzaiThePenguin: From what I can gather online, this is automatically disabled by the OS if view sound effects are disabled globally. So it shouldn't need to check for any preferences here.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
```diff
@@ -385,37 +397,125 @@ private void hideSystemBars()
...
-            // D-pad buttons are not holdable
-            if( DPAD.equals( assetName ) )
-                index = -1;
+            if( AppData.IS_KITKAT && mUserPrefs.isImmersiveModeEnabled )
+            {
+                // ignore edge swipes.
+                // unfortunately KitKat lacks a way to do this on its own,
+                // so just ignore all touches along the edges.
+                View view = getWindow().getDecorView();
+                if ( y < 10 || y > view.getHeight() - 10 || x < 10 || x > view.getWidth() - 10 )
```
BonzaiThePenguin: http://stackoverflow.com/questions/20530333/ignore-immersive-mode-swipe

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
littleguy77: You may have noticed that the current UI with the sliders does allow you to increment by 1% by tapping the +/- keys on either side of the slider.  Would be nice if we could let the user nudge with the new drag system.  Not sure how though without adding a lot of complexity.  Open to ideas...

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
littleguy77: I recommend you add the link inside the code comment as well.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
BonzaiThePenguin: Oh, I never bothered clicking those things. I'll change it to 1%.

BonzaiThePenguin:
Broken link since it says plus**e**, but I found it. :)

src/paulscode/android/mupen64plusae/input/map/TouchMap.java
littleguy77: Take a peek at the fields of the paulscode.android.mupen64plusae.util.Image class.  You might be able to just return `buttonMasks.get( i ).drawRect` or `buttonMasks.get( i ).drawable.getBounds()`.

littleguy77: This would work for any Image instance, buttons or analog.

I'll take a peek at the demapping... there may already be a static readonly map somewhere for this need.  If not, maybe it would be good to do it with a map for parallelism with the other code.  I'll have to refresh my memory... it's been a long time since I looked at this part of the code.

BonzaiThePenguin: Will do. The main problem here was that getButtonPress loops over the buttons to see which one intersects the touch input, but then it discards that information in favor of returning the N64 button index associated with the color written to what I assume is an off-screen image. I think I originally had it returning one of those rects, but by the time I figured out how to map the return value back to buttonMasks I had long since abandoned that code.

littleguy77: Ok, yeah this is where I forget most of the inner workings.

littleguy77:
This looks great!  I like that the popup is still accessible for fine-tuning and setting auto-holds (I just assumed you had removed it when I first read the PR).

I plan to merge this but just wanted to give it the thoughtful review it deserves.  Unfortunately my brain now mush after a long day, so it will have to wait until tomorrow.  If there's anything you want to rebase/squash/format in the meantime, go for it.

BonzaiThePenguin:
71c7faa fixes https://github.com/mupen64plus-ae/mupen64plus-ae/issues/416, and was accidentally included here but whatevs.

src/paulscode/android/mupen64plusae/input/map/TouchMap.java
BonzaiThePenguin: I changed it to return drawRect, after confirming that it was identical to the Rect I was creating manually. Overlooked that in my quest to figure out why the index couldn't be mapped to buttonMasks.

src/paulscode/android/mupen64plusae/profile/TouchscreenProfileActivity.java
BonzaiThePenguin: Confirmed that the sound is correctly disabled when system sounds are turned off. Acts exactly like a regular button.

PR #426, opened 2015-03-15
Implemented the navigation drawer

BonzaiThePenguin:
- MenuListView now supports disabled menu items, keyboard selection,
and checkable submenus
- Implemented the hierarchical navigation drawer using MenuListView


src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -214,7 +261,7 @@ public View getGroupView( int groupPosition, boolean isExpanded, View convertVie
                     Context.LAYOUT_INFLATER_SERVICE );
             View view = convertView;
             if( view == null )
-                view = inflater.inflate( R.layout.list_item_two_text_icon, null );
+                view = inflater.inflate( R.layout.list_item_menu, null );
```
BonzaiThePenguin: We can probably merge these and just ignore the extra ImageView on the right side when it's not needed.

src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -222,10 +269,23 @@ public View getGroupView( int groupPosition, boolean isExpanded, View convertVie
...
                 TextView text2 = (TextView) view.findViewById( R.id.text2 );
                 ImageView icon = (ImageView) view.findViewById( R.id.icon );
+                ImageView indicator = (ImageView) view.findViewById( R.id.indicator );
                 
                 text1.setText( item.getTitle() );
                 text2.setVisibility( View.GONE );
                 icon.setImageDrawable( item.getIcon() );
+                
+                if ( item.isChecked() )
+                    view.setBackgroundColor( 0x44FFFFFF );
```
BonzaiThePenguin: Transparent background color allows the keyboard selection color to show through.

src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -222,10 +269,23 @@ public View getGroupView( int groupPosition, boolean isExpanded, View convertVie
...
+                else
+                    view.setBackgroundColor( 0x0 );
+                
+                if ( item.getSubMenu() == null )
```
BonzaiThePenguin: Added support for collapse/expand arrows, which uses this new ImageView on the right side.

src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -142,7 +173,7 @@ public MenuListAdapter( MenuListView listView, Menu listData )
         @Override
         public boolean isChildSelectable( int groupPosition, int childPosition )
         {
-            return true;
+            return getChild( groupPosition, childPosition ).isEnabled();
```
BonzaiThePenguin: This functionality isn't actually used anywhere, but why not.

src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -64,18 +76,22 @@ public void setMenuResource( int menuResource )
...
         
-        // In case we want to add custom expand/collapse views to the groups
+        // Update the expand/collapse group indicators as needed
         setOnGroupExpandListener( new OnGroupExpandListener()
         {
             @Override
             public void onGroupExpand( int groupPosition )
             {
-                
+                reload();
```
BonzaiThePenguin: Reloading updates the collapse/expand arrows, which is why this line wasn't needed before.

src/paulscode/android/mupen64plusae/GalleryActivity.java
```diff
@@ -177,7 +181,7 @@ protected void onCreate( Bundle savedInstanceState )
         
         // Configure the navigation drawer
         mDrawerLayout = (DrawerLayout) findViewById( R.id.drawerLayout );
-        mDrawerToggle = new ActionBarDrawerToggle( this, mDrawerLayout, toolbar, R.string.app_name, R.string.app_name )
+        mDrawerToggle = new ActionBarDrawerToggle( this, mDrawerLayout, toolbar, 0, 0 )
```
BonzaiThePenguin: Accidentally had the label for the menu button set to the app's name, when being blank should be good enough.

res/values/strings.xml
```diff
@@ -62,6 +62,7 @@
     <string name="menuItem_refreshRoms">Refresh ROMs</string>
     <string name="menuItem_settings">Settings</string>
     <string name="menuItem_globalSettings">Global settings</string>
+    <string name="menuItem_profiles">Profiles</string>
```
BonzaiThePenguin: Only new string needed for the navigation drawer was a category for the different profiles.

res/menu/gallery_drawer.xml
```diff
@@ -3,14 +3,26 @@
...
+        android:icon="@drawable/ic_controller"
+        android:title="@string/categoryLibrary_title"/>
+
+    <item
         android:id="@+id/menuItem_settings"
         android:icon="@drawable/ic_settings"
-        android:title="@string/menuItem_settings">
+        android:title="@string/menuItem_settings"/>
+
+    <item
```
BonzaiThePenguin: Language below the Settings should be good enough since it has the Map icon, and Settings will be accessed much more often.

res/layout/list_item_menu.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BonzaiThePenguin: This is identical to the standard list item, but it has an ImageView on the right side. We can probably merge these.

src/paulscode/android/mupen64plusae/MenuListView.java
```diff
@@ -180,10 +211,26 @@ public View getChildView( int groupPosition, final int childPosition, boolean is
...
+                
+                view.setBackgroundColor( 0x0 );
+                
+                // Indent child views by 15 points
+                DisplayMetrics metrics = new DisplayMetrics();
+                ((Activity) mListView.getContext()).getWindowManager().getDefaultDisplay().getMetrics(metrics);
+                
+                view.setPadding( (int) ( 15 * metrics.density ), view.getPaddingTop(), view.getPaddingRight(), view.getPaddingBottom() );
+                
+                if ( !item.isCheckable() )
```
BonzaiThePenguin: Added the ability for child rows to have checkboxes using the ImageView on the right side.

res/layout/gallery_activity.xml
```diff
@@ -1,17 +1,20 @@
...
+<?xml version="1.0" encoding="utf-8"?>
+<android.support.v4.widget.DrawerLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:orientation="vertical"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"
+    android:id="@+id/drawerLayout"
+    android:fitsSystemWindows="true" >
```
BonzaiThePenguin: Updated the layout so the navigation drawer appears above the Toolbar, which is the recommended behavior according to Google. This required some changes to the view hierarchy in addition to this flag.

res/layout/gallery_activity.xml
```diff
@@ -21,21 +24,24 @@
...
 
-        <!-- navigation drawer -->
-        <RelativeLayout
-            android:layout_gravity="left"
-            android:layout_width="260dip"
-            android:layout_height="match_parent">
+    <!-- navigation drawer -->
+    <RelativeLayout
+        android:layout_gravity="left|start"
+        android:layout_width="280dip"
```
BonzaiThePenguin: Also made the navigation drawer 280 units wide instead of 260 units.

littleguy77:
Excellent, thanks!  Would you mind if I auto-format these changes?  Or will that throw off your next PR?  The only reason I is because it often makes diffs, merges, cherry-picks go more smoothly later.  Otherwise, I can just do it at the very end.

res/layout/gallery_activity.xml
littleguy77: Yeah I was wondering about that.  I actually kind of like the spinning bars/arrow animation, but if that's what they recommend, whatever.

res/values/strings.xml
littleguy77: Thanks

res/menu/gallery_drawer.xml
littleguy77: Definitely.  I'm happy pushing this down as far as you want, even to the bottom TBH.

res/layout/gallery_activity.xml
littleguy77: Actually this makes sense since it maximizes usable space in the drawer.

PR #440, opened 2015-05-31
Add dependencies for GLideN64/GLideNHQ

littleguy77:
I wanted to integrate the dependencies first as a separate set of commits, since they affect other modules as well (e.g. freetype for mupen64plus-core and libpng for rice and glide4mk2).

@Gillou68310 @gonetz I'm opening a pull request for discussion before merging this, just to be sure I haven't misunderstood any of your discussions.  Please let me know if this is indeed how you want to proceed with these dependencies (e.g. are we in agreement on versions?).


gonetz:
That work was done entirely by Gillou68310. My work uses his work.
The only notice: I would add FreeType includes to Android.mk right into dd60c03

littleguy77:
@gonetz Good idea, done.

Gillou68310:
@littleguy77 I updated libpng to 1.6.10 to see if it helps with the crash @gonetz had on some devices. It seems like these crashes were caused by corrupted png files so if gliden64's code is still compatible with older png library we're not forced to update it.

We should be carefull not to include both "src/inc" and "src/GLideNHQ/inc/" in gliden64 and glideNHQ makefiles in order to avoid conflicts between library versions.

ae-imports and SDL2 shared library are not necessary for gliden64 https://github.com/Gillou68310/mupen64plus-ae/commit/bb3ee674b688b2b4211db94f6fadbe9691e5611c

I guess this one https://github.com/Gillou68310/mupen64plus-ae/commit/37786f720aad9c216d1fafb418740f9f6c6b425f could be merged directly to master.

gonetz:
I already updated GLideNHQ to work with new libpng, and this update is in master.
It worth to do because I had a problem on Linux: headers from old libpng, binaries from the new one.
Now everything is consistent.

Gillou68310:
Sure but we're not using the provided libpng headers on android so we can use any compatible version of libpng. I just checked and there is no build issue with the currently used libpng (1.4.6)

littleguy77:
@Gillou68310 is correct.  We are not using the libpng headers or binaries provided in the GLideN64 repository.  We can keep the version we had been using.  I just tested Super Mario with hi-res textures on a GLES2 device and had no issues.

gonetz:
Ok, so my corrections are backward compatible :)

littleguy77:
I just rebased.  This PR contains only GLES3 and FreeType, and is sufficient for GLideN64/HQ integration.  @Gillou68310 if you have no issues, please click the merge button.

gonetz:
is 3.1 branch still available? I need 3.1 build for debugging.

littleguy77:
Yes, GLES 3.1 is included.  My latest branch should have all the fixes from your recent comments on the forum:
https://github.com/littleguy77/mupen64plus-ae/commits/gliden64-integration

- fixed pull script
- removed unnecessary dependencies in make files
- corrected the displayed config options

gonetz:
Great, thanks! I just found that auxiliary frame buffers do not emulated properly, regardless of GLES level. Can't find, what's wrong yet.

PR #441, opened 2015-06-01
video-gln64: Remove ae-bridge dependency

Gillou68310:


littleguy77:
Thanks that always bothered me :+1:

PR #442, opened 2015-06-06
Gliden64 integration

littleguy77:
@Gillou68310 @gonetz
Any reason not to merge this into master now?  I mainly ask is because we'll probably see an increase in test/bug reports once it's in master and being auto-built.  Don't want to overload everyone if there's still core stuff you want to focus on.


Gillou68310:
I was wondering if there's a way to detect the major version of gles on the java side? So people with gles2 device cannot select gles3 plugin. This should save a few annoying complains ;-)

littleguy77:
Yeah, the recommended way by google is to simply request a context and see what you get back:
http://developer.android.com/guide/topics/graphics/opengl.html#version-check

Definitely something to get in before we publish but for now I'm ok letting it go while we focus on bigger issues on both the native and front-end side.

Gillou68310:
Ok then it's fine for me too :-)

gonetz:
I have no objections. Besides, Lars Bishop's patch for TGD is based on latest commit in this chain.

PR #446, opened 2015-08-12
Pull latest GLideN64 and fixed default profile issue, also now including dynamic extraction of zip files

fzurita:
I went ahead and pulled the latest GLideN64, I did not mean to initially include this in my pull request but I'm not sure how to back it out. Hopefully this is not an issue. It seems to be working better than the one in the current build though.

I also fixed an issue where the user selected default profiles are not being selected correctly when first entering a game if one hasn't visited the GamePrefsActivity at least once.


fzurita:
I found a couple of bugs with zip extraction, closing this until I fix them.

PR #451, opened 2015-08-21
Load the correct default profile if we have not yet visited the  GamePrefsActivity for a game.

fzurita:
Currently, if one loads a game without first going to the GamePrefsActivity at least once for a specific game. The game will not load the correct default profile. This fixes that.


littleguy77:
Thanks!

PR #497, opened 2015-10-23
Tv mode improvements

fzurita:
Fixed all known regressions in original pull request


Gillou68310:
Nice work!
I'm away from my computer so I won't be able to test and merge this until tuesday, sorry for the delay.

I noticed that the action bar is missing from the manage profile activity on master branch, it might be related to the last 2 PR I merged , could you take a look at it when you have some time.

Gillou68310:
Also I found an odd combo with the gameside bar the causes a crash of the UI. You have to try opening the gameside bar with 3 fingers than try to slide down the notification bar with a 4th finger and then release everything at the same time.

Here's a screenshot:
screenshot_2015-10-24-11-19-25

fzurita:
Wow, interesting combo. Had this been around before the controller accessibility changes?

fzurita:
The action bar is present on the new TV mode improvements branch rebased with master. Once we merge that in, it will come back.

Gillou68310:
Ok everything looks good here.
I just have one comment, the radio buttons in the select slot popup are almost invisible, is it possible to have white radio buttons?
@littleguy77 is this ok for you too?

fzurita:
I can change the theme, they look visible in my device

Gillou68310:
Hum I tried to make a screenshot but radio buttons are a lot more visible on the screenshot than on the real device. You have to trust me on this one ;-)

Gillou68310:
> Edit: I can't reproduce it, can you grab a logcat?

[crash.txt](https://github.com/mupen64plus-ae/mupen64plus-ae/files/21586/crash.txt)

This is not looking good. Also this is happening on both master and this branch, so no need to fix it in this PR

fzurita:
Wow, logcat is not very helpful. It's not telling us what in our code is causing the crash. I'll google and see if I can find something.

fzurita:
Well... google is telling me that it could be an issue with the android support library itself....

Gillou68310:
I told you that this was not looking good :-)

fzurita:
There seem to be some hacky workarounds. I'll do some more research...

fzurita:
The solution for the hard to see radio buttons is not difficult. See here:

http://stackoverflow.com/questions/26843098/android-appcompat-v721-0-0-change-material-checkbox-colors

If that is the only thing preventing the pull request from getting merged, I'll fix that tonight.

fzurita:
So the whole black radio button issue seems to be a bug in the android support library....

fzurita:
And I think I found a solution. I'll add it to the pull request.

Gillou68310:
Good job ;-)

fzurita:
Feel free to merge all the other pull requests as well. :)

PR #498, opened 2015-10-23
Cleaned up warnings

fzurita:


fzurita:
Yeah, I left them commented because I thought we may actually need them in the future, but I just wanted to get rid of the warnings.

fzurita:
Ok, all cleaned up.

Gillou68310:
This branch has conflicts you need to rebase against master

PR #499, opened 2015-10-24
Fix incorrect gallery art

fzurita:
Fixed issue where the incorrect gallery art is shown momentarily if scrolling very fast.

### DIFF
--- a/src/paulscode/android/mupen64plusae/GalleryItem.java
+++ b/src/paulscode/android/mupen64plusae/GalleryItem.java
@@ -141,6 +141,7 @@ public class GalleryItem
     {
         public GalleryItem item;
         private Context mContext;
+        LoadBitmapTask mLoadBitmapTask = null;
         
         public ViewHolder( Context context, View view )
         {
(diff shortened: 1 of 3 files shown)


Gillou68310:
> Fixed issue where the incorrect gallery art is shown momentarily if scrolling very fast.

This is still happening on my device.

fzurita:
Hmmm, not good, I may have to try this on other devices. It worked OK on the onr device I tested

fzurita:
Well, I think I actually know what the problem is. The way I'm cancelling the ASync task is not reliable enough. I can make additional commits to fix it. It just requires one additional change in the task it self so that it is aware that it was cancelled.

fzurita:
Ok, I made a new commit. Please try again when you get a chance.

Gillou68310:
Nope still happening.

fzurita:
Hmmm, I wonder what's different in your device.

fzurita:
What version of Android do you have?

Gillou68310:
4.4.2

fzurita:
I'm testing on devices with 5.1. I actually have a device with that version of android, I'll see if I can reproduce it there.

Gillou68310:
Ok ;-)

fzurita:
So, I just checked this out in my 4.4 device and it seems to be working correctly there. Just to make sure, the default covert art is supposed to show up while the real cover art is loading.

Before this fix, the cover art for another game could show while the real cover art was loading. Is that what you are seeing in your device with this fix?

Gillou68310:
Oh sorry for the misunderstanding I thought the default covert was the incorrect gallery art.

PR #500, opened 2015-10-25
Profile improvements

fzurita:
Fixed issue that caused the wrong profile to be selected if a custom profile is currently default.

Also, added a "Current Default" profile that will always select the current default

Made this profile the default profile.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -416,6 +416,7 @@
     <string name="controllerProfile2_title">Controller 2 profile</string>
     <string name="controllerProfile3_title">Controller 3 profile</string>
     <string name="controllerProfile4_title">Controller 4 profile</string>
+    <string name="default_profile_title">Current Default</string>
     
     <!-- Other Preferences -->
     <string name="GameAutoSavesMax_title">Max auto saves per game</string>
--- a/src/paulscode/android/mupen64plusae/persistent/GamePrefs.java
+++ b/src/paulscode/android/mupen64plusae/persistent/GamePrefs.java
@@ -355,17 +355,17 @@ public class GamePrefs
     
     private static Profile loadProfile( SharedPreferences prefs, String key, String defaultName,
         ConfigFile custom, ConfigFile builtin )
-    {
+    {        
         final String name = prefs.getString( key, defaultName );
 
         if( TextUtils.isEmpty( name ) )
             return null;
         else if( custom.keySet().contains( name ) )
             return new Profile( false, custom.get( name ) );
-        else if( custom.keySet().contains( defaultName ) )
-            return new Profile( false, custom.get( defaultName ) );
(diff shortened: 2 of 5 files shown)


Gillou68310:
"Current Default" profile is missing for Controller profile

fzurita:
You have to select a default profile first.

fzurita:
Although, I could probably allow it without a default profile selected.

Gillou68310:
I guess (Disabled) should be the default profile, right now there is a %1$s instead of the default profile

fzurita:
Where do you see the " %1$s" text?

Gillou68310:
In game preference activity, under Controler profile.

fzurita:
Hmmm... not good, I didn't see it in my testing. I'll check it out tonight.

Gillou68310:
Ok

Gillou68310:
Any news about this, this is still an issue by my side. Just uninstalling - re-installing you should be able to reproduce it

fzurita:
I didn't really have much time last night, I'll probabky get to it tonight.

Gillou68310:
Ok no problem just wanted to be sure you didn't forget about it ;-)

fzurita:
Nope, I haven't, it's in my TODO list.

fzurita:
Just to clarify, it only happens in the Controller profiles?

fzurita:
I think to fix it, I'm going to add a default profile of "None" to the controller profiles. Do we have some code somewhere that selects the correct controller profile based on the current controller? I will have to only set the default to none if no controller is detected if that's the case.

Gillou68310:
> Just to clarify, it only happens in the Controller profile selection?

Yep

littleguy77:
Yeah I bet it happens because there is no global default controller profile.

I think the default for the Controller Profile value in each game should be "Disabled" rather than "Current Default" (and adding a "None" profile to the built-ins).  Trying to auto-detect the plugged controller is a nasty rabbit hole that is best to avoid.  Too many variations between controllers, Android devices, and Android versions to do it well.

fzurita:
I think I will keep the "Current Default" option. The "Current Default" option should correctly point to the "None" controller profile.

fzurita:
How about a compromise. How about we keep the "Current Default" option for controllers and make "Disabled" the default rather than a "none" built in profile?

littleguy77:
Sounds good to me.  That's basically what I was saying.  I like the Current Default option.  I just thought a None profile was a little hackish.

fzurita:
Gotcha

littleguy77:
Minor note, you might consider calling it Global Default rather than Current Default..  Maybe a little more clear?  Just a thought.

fzurita:
Not a bad idea, I'll do that as well.

PR #505, opened 2015-10-31
Reset function now functions correctly.

fzurita:


fzurita:
Hmm.... one of my previous commits broke controller navigation of the gameSidebar.... I'll have to take care of that.

PR #509, opened 2015-11-02
Mupen64plus core interaction enhancements and UI enhancements

fzurita:
* Gameplay will be restored automatically and the sidebar closed if a slot or file is loaded or saved. This involved fixing a lot of corner cases.

* The gallery game side bar is now closed automatically when launching a game. This forces a refresh of the gallery sidebar so the the resume button shows up when a game is opened for the first time.

* Now no longer showing the toast staying that a game is being resumed when a game is being restarted. This is done correctly this time.

* We will now no longer exit the GameActivity immediately after a user chooses to exit. We will now wait until the core is done fully shutting down.

* Added a reset option to the in-game sidebar.

### DIFF
--- a/res/menu/game_drawer.xml
+++ b/res/menu/game_drawer.xml
@@ -69,5 +69,8 @@
     <item
         android:id="@+id/menuItem_setIme"
         android:title="@string/menuItem_setIme"/>
+    <item
+        android:id="@+id/menuItem_reset"
+        android:title="@string/inputMapActivity_reset"/>
 
 </menu>
--- a/src/paulscode/android/mupen64plusae/GalleryActivity.java
+++ b/src/paulscode/android/mupen64plusae/GalleryActivity.java
@@ -32,9 +32,7 @@ import java.util.Locale;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
-
 import org.mupen64plusae.v3.alpha.R;
-
 import paulscode.android.mupen64plusae.GameSidebar.GameSidebarActionHandler;
 import paulscode.android.mupen64plusae.dialog.ChangeLog;
 import paulscode.android.mupen64plusae.dialog.Popups;
@@ -57,6 +55,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.Configuration;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.view.GravityCompat;
@@ -77,6 +76,7 @@ import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.ViewTreeObserver;
 import android.view.inputmethod.InputMethodManager;
 
 public class GalleryActivity extends AppCompatActivity implements GameSidebarActionHandler
@@ -85,6 +85,7 @@ public class GalleryActivity extends AppCompatActivity implements GameSidebarAct
     public static final String STATE_QUERY = "query";
     public static final String STATE_SIDEBAR = "sidebar";
     public static final String STATE_CACHE_ROM_INFO_FRAGMENT= "cache_rom_info_fragment";
+    public static final String STATE_GALLERY_REFRESH_NEEDED= "gallery_refresh_needed";
     
     // App data and user preferences
     private AppData mAppData = null;
@@ -117,6 +118,9 @@ public class GalleryActivity extends AppCompatActivity implements GameSidebarAct
     
     //True if the restart promp is enabled
     boolean mRestartPromptEnabled = true;
+    
+    //If this is set to true, the gallery will be refreshed next time this activity is resumed
+    boolean mRefreshNeeded = false;
     
     @Override
     protected void onNewIntent( Intent intent )
@@ -312,6 +316,8 @@ public class GalleryActivity extends AppCompatActivity implements GameSidebarAct
             String query = savedInstanceState.getString( STATE_QUERY );
             if( query != null )
                 mSearchQuery = query;
+            
+            mRefreshNeeded = savedInstanceState.getBoolean(STATE_GALLERY_REFRESH_NEEDED);
         }
         
         // find the retained fragment on activity restarts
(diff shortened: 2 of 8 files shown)


Gillou68310:
I noticed some wrong cases:
- If selecting the exit item then canceling on prompt confirmation, game resumes while game sidebar is still opened
- If game sidebar is opened and home button is pressed, restoring app will resume game while sidebar is still opened

 I think the sidebar should also be closed if saving a slot or file, what do you think?

fzurita:
Nice testing. Thanks for finding those cases.

Also, yeah, it probably makes sense to close the side bar in those cases as well.

fzurita:
I see the problem with with the emulator resuming when the home button is hit and then we go back to the app. The ```GameSurface.createGLContext()``` we resume the emulator if it's currently not in the running state.

This also means that if the user pauses the emulator using an assigned controller key, if they hit the home button and come back to the app, the emulator will come back as unpaused. I can't think of any other case where it will break.

fzurita:
This should be good to go now, I'm going to look at the crashing and save state not loading issues when moving the toast next.

Gillou68310:
Still one issue, if you press the sleep button and sidebar is closed, when waking up the phone the emulation doesn't resume.

Also doing an advanceFrame() when surface is recreated is a good idea but it doesn't seems to work for every game I tested.

Another option is to save the framebuffer to texture and draw that texture when egl surface is recreated, but that's a lot of work for nothing IMO.

The easiest solution for now would be to just close the sidebar (and possibly opened dialog) and resume the game, what do you think?

fzurita:
What game for advance frame not work with? Maybe advancing 2 frames will work with all games.

fzurita:
Hmm... redrawing the menu when gallary comes back up after exiting a game doesn't quite work because the save hasn't happened yet.

Gillou68310:
> What game does advance frame not work with? Maybe advancing 2 frames will work with all games.

Rayman2 and I tried advancing 3 frames it still doesn't work :-(

> The only easy solution I can think of is to add an Async task that just waits 1 second and then checks for the presence of any auto saves after exiting a game. The other option is to just assume that an auto save will be present after exiting a game.

Maybe you can force the sidebar to close only on game exit?

fzurita:
Yeah, that's probably a better option. I can close the sidebar if a game is launched either through re-start or resume.

Gillou68310:
Just make sure the closing is done after the game surface is created, just to be sure it's not visible to the user.

fzurita:
I had a segfault from the core when loading Rayman 2 for the first time. See here:

https://drive.google.com/file/d/0B57Ioy26LWegQlpmQnV1MDJ0VnM/view?usp=sharing

Gillou68310:
You mean when exiting rayman for the first time?

fzurita:
Yes, it did happen while trying to exit.

Gillou68310:
Is this still happening?

fzurita:
It only happened the first time I tried to load then quickly exit the game.

fzurita:
I didn't let it finish loading though.

Gillou68310:
Your log is weird because the egl context is created 2 times I don't understand how this happens!

Gillou68310:
Does it happens on this branch or master?

fzurita:
Just this branch.

Gillou68310:
Maybe a race condition with the game surface listener

Gillou68310:
Do some more testing and just see if it happens again

fzurita:
Try opening a game that you haven't opened before. And before it's done loading, exit the game. That's when this happened to me.

Gillou68310:
I think this is what caused the crash:
11-06 11:33:27.326: I/GameLifecycleHandler(31347): onCreate

How on earth did the GameLifecycleHandler got re-created?

fzurita:
It will be re-created each time a game is launched.

Gillou68310:
Sure but not while the game is already running!

fzurita:
I think I know what's happening. It's because the Core runs in another thread. So if I run a game and exit it, which brings me back to the gallery, but the core is still running in the background. Then I very quickly launch another game before the core is done exiting....

fzurita:
I don't think the CoreInterface.shutdownEmulator() is really waiting for the core to shut down before showing the gallery again.

Gillou68310:
No it's not I already noticed that

Gillou68310:
Maybe we should wait until core is finished before calling a finish activity

fzurita:
And this is happening because the CoreInterface.exit() call is killing the GameActivity. The GameLifecycleHandler.onDestroy method is being called while the Gallery is up in the background.

Gillou68310:
Yep ;-)

Gillou68310:
That was intended by the way

fzurita:
The easy fix is to only call sActivity.finish in the shutdownEmulator call. I would really just combine the CoreInterface.exit and CoreInterface.shutdownEmulator methods. Then nothing at all would get called in onDestroy.

Gillou68310:
I did that at some point but reverted it I can't remember why.
Ok let's do this and see what's happening

fzurita:
I think I know why. It would add a considerable delay when exiting the emulator. Maybe the right answer is to make sure that the core is still not running when launching a new game.

fzurita:
Hmm... I don't get a black screen in Rayman 2 when going home and coming back to the app in my 4.4 device either. Can you check if the issue is still happening to you with the newest version of this branch?

Gillou68310:
> I think I know why. It would add a considerable delay when exiting the emulator. Maybe the right answer is to make sure that the core is still not running when launching a new game.

Sounds like a good idea but that would required communication between gallery activity and game activity like you said earlier.

> Hmm... I don't get a black screen in Rayman 2 when going home and coming back to the app in my 4.4 device either. Can you check if the issue is still happening to you with the newest version of this branch?

Give me a sec

Gillou68310:
Ok black screen only occurs if framebuffer emulation is disable, so it may not work with other video plugins as well (haven't checked)

If you're in portrait mode, start a game in landscape mode then exit the game while still in landscape mode the gallery sidebar isn't closing.

Going back home I won't be able to do more tests until tomorrow ;-)

fzurita:
You are good at finding all the corner cases.

Gillou68310:
They usually find me first ;-)

> I think I know why. It would add a considerable delay when exiting the emulator. Maybe the right answer is to make sure that the core is still not running when launching a new game.

Just got an idea, doing a finish activity when core thread  is finished should possibly fix the swap buffer exception you get on your device. But surely will add a delay when exiting :-(

fzurita:
I agree, it will fix that. It is also going to add a good amount of delay when exiting a game.

fzurita:
I just realized that all the toast messages in CoreInterface.startupEmulator will crash the app because they are all being called outside the main UI thread

fzurita:
I went ahead and took a crack at it. I don't really see much difference in how long it takes to exit.

littleguy77:
Wow... most github notifications in a day LOL

Glad to see I'm not the only one who pulled his hair out trying to figure out the lifecycle mechanics!  The number of corner cases is astonishing.  There are a bazillion use case sequences.  Add to that the variations between different Android versions.  This is why I always got nervous when @Gillou68310 brought up the "don't destroy GL context" branch... ;)  Took me forever just to get something that didn't crash.

I am so far out of this particular conversation, so I won't try to steer you in any way.  Just wanted to share some of my experiences.  IIRC, I eventually found that the only way I could "win" was to treat the Android lifecycle as master, and slave the OpenGL and core lifecycles to it.  Whenever I viewed the core's lifecycle as the master, things usually seemed to spiral down a rabbit hole.  Patch a corner case only to reveal another...

fzurita:
I know... Even a minor change in this area of the code seems to break something else. I think we are getting there though.

littleguy77:
Cool.  Quick question, is the purpose of this PR still "Opening the in-game sidebar now pauses gameplay", or has it evolved to do more?  Just wondering because sounds like a lot of effort and changes for a seemingly minor UX enhancement.

fzurita:
It really has evolved significantly. I'm sure you gather that by the conversation. Just to get this working correctly has meant a lot of changes.

littleguy77:
Yes, the implementation has evolved significantly.  What I'm unclear on is whether the *feature* that this PR implements is still merely to pause the emulator when sidebar is open.  If this PR now implements several new features besides just pausing the emulator, please update the title or first comment at the top of this page.  Thanks ;)

fzurita:
Yep, of course :)

Gillou68310:
> Still one issue, if you press the sleep button and sidebar is closed, when waking up the phone the emulation doesn't resume.

This is still an issue.

Also I suggest adding a dialog when selecting the "reset" option, so users don't lose their progress if accidentaly touching the option.

Everything else look good, nice work ;-)

Gillou68310:
We still need to do something about the advance frame thing, but this is not a functionnal issue so it can be done later.

fzurita:
Hmmm, I thought I fixed that... Looks like I broke it again. I think I know what's wrong though. I'll also add the confirmation dialog.

fzurita:
Which plugin had the advanced frame issue?

Gillou68310:
gliden64 has it with framebuffer emulation off, I haven't checked the others

xperia64:
I just tried the latest autobuild from 11-04 and the orientation option seems to be missing.

fzurita:
Hmmm, interesting. Where s this option located? I haven't really messed around much with it.

fzurita:
I see the option under app settings.

xperia64:
Oh yep it's there. I just was used to it being next to the resolution settings which are now per-game.

fzurita:
Ok. The restart confirmation dialog has been added. Also I again fixed the case where we would not resume correctly if a device was suspended.

fzurita:
This branch has gotten really messy in the commit history. Before you merge it, I'm going change history to make it neater.

fzurita:
Ok, I'm done redoing the commit history.

PR #511, opened 2015-11-10
GLideN64 update nov 9

fzurita:


Gillou68310:
Yes I do, copyAuxToRDRAM is not necessary for every games and should be used only when required.

fzurita:
Ok, I'll make that the default

fzurita:
Ok, it is now false by default.

PR #512, opened 2015-11-10
Now saving game state when the home button is hit.

fzurita:
Currently, if a user goes home and then kills the app or the OS kills the app, the game state will not be saved.

This fixes that issue. Also, one oddity, for some reason the Core refuses to save the game unless it's resumed for at least 100ms.

### DIFF
--- a/src/paulscode/android/mupen64plusae/jni/CoreInterface.java
+++ b/src/paulscode/android/mupen64plusae/jni/CoreInterface.java
@@ -385,24 +385,37 @@ public class CoreInterface
         }
     }
     
-    public static synchronized void pauseEmulator( boolean autoSave, String latestSave )
+    public static synchronized void pauseEmulator(boolean autoSave, String latestSave)
     {
-        if( sCoreThread != null )
+        if (sCoreThread != null)
         {
-            NativeExports.emuPause();
-            
-            // Auto-save in case device doesn't resume properly (e.g. OS kills process, battery dies, etc.)
-            if( autoSave && latestSave != null)
-            {                
+
+            // Auto-save in case device doesn't resume properly (e.g. OS kills
+            // process, battery dies, etc.)
+            if (autoSave && latestSave != null)
+            {
+                NativeExports.emuSaveFile(latestSave);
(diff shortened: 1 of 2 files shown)


src/paulscode/android/mupen64plusae/jni/CoreInterface.java
```diff
@@ -385,24 +385,37 @@ public static synchronized void resumeEmulator()
         }
     }
     
-    public static synchronized void pauseEmulator( boolean autoSave, String latestSave )
+    public static synchronized void pauseEmulator(boolean autoSave, String latestSave)
     {
-        if( sCoreThread != null )
+        if (sCoreThread != null)
```
Trass3r: I would have also inverted the if for early exit.

Gillou68310:
This is happening because the core is already paused when the savestate command arrives. I was wondering if it's possible to detect that the app is going to background and only pause the core after the savestate cammande has been sent?

fzurita:
The problem is that someone else is pausing the core before the court is paused here.

fzurita:
Addressing your previous comment. It is certainly possible to detect when the app is going to the background. That is what "onStop" is doing. The problem is that if someone else pauses the core before onStop gets called, for example if the user pauses the game by opening the menu. If the core is already paused, we are going to have to resume momentarily to save the state.

I think this is the correct solution for that case.

Gillou68310:
onStop is also called when pressing the sleep button and autosave is not necessary in this case.
I would rather do this in onSurfaceDestroyed.
Also I think this should be done in a seperated method (or directly in onSurfaceDestroyed) because pauseEmulator with autosave is also called when exiting the core and resuming for 100ms isn't necessary in this case.

fzurita:
What if the user has the game paused (side bar open ) when they hit the home button?

Gillou68310:
> What if the user has the game paused (side bar open ) when they hit the home button? I think we would still have to momentarily resume the game.

Sure

fzurita:
This is not really needed. I'm going to go ahead and close it.

PR #515, opened 2015-11-26
Keep dialogs on screen rotation and new support library

fzurita:
Converted most used dialogs into dialog fragments so that screen rotation doesn't dismiss them.

Because Preference Dialogs don't support Fragments, I had to import the new Preference support library as well as update to the latest support library. This cascaded into a lot of changes. But on the plus side, we are no longer using deprecated preference activities.

There are two things I'm unable to test:
* One of the changes required me to switch the Xperia game activity from a NativeActivity to a regular activity. We need to make sure that the xperia play still works.
* I added three new support library related projects. Will this break the auto build? If so, how do we fix it.

Dialogs that have changes:
* All preference derived dialogs
* Profile related dialogs
* Confirmation dialogs
* Controller input related dialogs
* Cheat editor related dialogs

I did not touch these dialogs, so they still dismiss on screen rotation:
* Dialogs in the Popus class.
* Remaining dialogs in the Promp class.
* Dialogs for extracting high resolution textures in rice plugin.

Also, all preferences are now using preference fragments instead of preference activities.

Screen rotation should now no longer affect dialog state for the affected dialogs.


littleguy77:
On this Thanksgiving day I'm so glad the project is in such capable hands @fzurita @Gillou68310 @xperia64 . :)

The Xperia Play device's special touchpad won't work if the activity doesn't derive from NativeActivity.  That said, this may be a good time to think about dropping support for it anyhow, and bumping the minimum SDK.  At this point the XP is an *ancient* device, and it's crazy how much of the codebase is dedicated to a single device.  Our backwards compatibility at this point is downright *noble*.

I recommend you open a post on the forum to discuss bumping minimum SDK to ICS or JB.  You could justify it using [these stats](http://developer.android.com/about/dashboards/index.html) and the fact that it would streamline a lot of the code relating to fragments and so forth.

Keep up the great work!

fzurita:
You are very right about the stats. Sounds like it's time to drop support for such old devices. It would definitely help us clean up the code even more.

fzurita:
Going to API  15 (ICS) would mean that we no longer need any support libraries for our uses.  And it wouldn't be hard to switch everything from the Android support library since most calls are equivalent.

fzurita:
Ok, I was totally wrong. In some places, the support libraries are used to add functionality in addition to backward compatibility.

Examples are DrawerLayout, RecyclerView, and PreferenceDialogFragment. So unless we decide to write or own versions of those classes, we can't get rid of the support libraries.

fzurita:
I'm not getting any feedback in this pull request. Also @paulscode is looking into making sure that the auto build script keeps working correctly.

Does anyone mind if I go ahead and merge this? Also, should I cut off xperia play support?

Gillou68310:
I probably won't have time to test this anyway soon so go ahead and merge it. Also considering the stats I see no reason to keep xperia play support!

fzurita:
I will drop xperia play support in a subsequent pull request.

PR #519, opened 2015-12-07
Rom Hacks not working (serious this time)

SnakeHugz:
I have recently tried to play a rom hack that I had patched with a patcher app. It's not working.  Help


fzurita:
Please write an issue instead. It is known that ROM hacks don't work.

Justanthony123:
The ROM is not working I was playing super Mario 64 and then I got to the end and now it will not let me by pass the end screen

fzurita:
Select the game and choose start instead of resume. In the real console, the only way to get past that screen is to reset the console.

Justanthony123:
What do you mean reset the console this is an app

fzurita:
As in you have to do the same thing within the app as the real console. Just hit the back button in your phone while within the game and choose exit.

After that, select the game and hit start instead of resume.

Justanthony123:
https://photos.app.goo.gl/ydbH5XzcvJ5sYJiz8 so this is a photo of the app and it does not say start anyway and when I hit play it take me back to the end screen.

fzurita:
Wow, you are using a very old version of the app. This is the latest version:

https://play.google.com/store/apps/details?id=org.mupen64plusae.v3.fzurita

PR #526, opened 2015-12-14
Immersive mode issue

fzurita:
Fix issues with immersive mode.

Fix issues with stretch mode not working correctly with GLideN64 at native resolution.

### DIFF
--- a/src/paulscode/android/mupen64plusae/game/GameActivity.java
+++ b/src/paulscode/android/mupen64plusae/game/GameActivity.java
@@ -214,6 +214,7 @@ OnPromptFinishedListener, OnSaveLoadListener, GameSurfaceCreatedListener, OnExit
         
         // Enable full-screen mode
         window.setFlags( LayoutParams.FLAG_FULLSCREEN, LayoutParams.FLAG_FULLSCREEN );
+        window.setFlags(LayoutParams.FLAG_LAYOUT_IN_SCREEN, LayoutParams.FLAG_LAYOUT_IN_SCREEN);
         
         // Keep screen from going to sleep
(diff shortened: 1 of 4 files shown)


Gillou68310:
The resolution in streched mode is wrong since this PR (at least on my device), I got 1440x1080 instead of 1920x1080

fzurita:
Interesting, so it's not getting stretched?

fzurita:
I just re-tested this and it's getting stretched correctly, at least for GLideN64 and Glide64.

You are right, the rendered resolution is 1440x1080, which sounds correct to me? Unless the plugin has some sort of wide screen hack that outputs at a different aspect ratio?

Gillou68310:
It's getting streched correctly, but widescreen hack isn't working anymore because of the resolution ratio. Also I'm not sure it's a good idea to let android strech the image because we can lose in image quality. I'm doing something similiar with the zoom option and I noticed that the image quality was impacted.

fzurita:
@Gillou68310 Looking at the older code, the lower resolutions actually stretch the raw output of the plugins at the lower resolutions. Only the native resolution didn't stretch the plugin output and actually caused the plugins to render at the proper stretched resolution.

fzurita:
Ok, I was able to fix the issue at native resolution. I did it under this pull request:

https://github.com/mupen64plus-ae/mupen64plus-ae/pull/528

It turns out that GLideN64 has no issues with stretch mode at native resolution. For some reason my emulation.cfg file had all AspectRatio values set to 1. I don't see anywhere in the code where we actually set that value. I'm now completely ignoring the AspectRatio value in the emulation.cfg file when creating the native config file.

PR #530, opened 2015-12-22
Add sensors to Controller profiles

javaxubuntu:
I made these changes because I wanted to use the sensor of my phone to emulate the analog stick.
When I developed it, it caused problems on game menus because is's difficult to have sensors exactly on position 0. So this sensor is disabled when the game is launched, and I created 2 new mappable buttons on Controller profiles: «enable sensor» and «disable sensor», that allow to enable/disable sensor during game.
There is (still) no change on Touchscreen. Maybe there should be a new toggle button to enable/disable sensors, and why not hiding analog stick when sensor is enabled, but I don't know what should be the best for other users.


fzurita:
Cool

fzurita:
I see that you added a bunch of ant build scripts. That's useful.

fzurita:
In my opinion, the Enable Sensor/Disable sensor buttons should be combined into 1 button. This button should toggle the sensor on/off. You could name it "Toggle Accelerometer" or "Toggle Sensor"

fzurita:
By the way, good job on this. I tried it out and it works great. Now here are a few thing that could be improved in this implementation.

Is there any reason you added this to the controller mappings instead of touchscreen mappings?

It would make sense to add it to the touch screen profiles since this is using a phone's accelerometer. Unless you have plans to make this work with controller accelerometers?

What you could do is add another draggable button to enable/disable the accelerometer as a joystick to the touchscreen controller. Also, I don't see any reason to use the accelerometer for anything other than joystick control so I would not allow mapping to anything other than that.

If you make the above change, I think we would be all set and we can merge it.

javaxubuntu:
I added this on the controller because it was easier. I added it to the touchscreen, but I still have to add a mapping that can at least disables Y axis and configure sensitivity. Also, I have to disable the SensorEventListener when it is not used.
Why not adding a button on the the controller mappings? Thus, the sensor is neither in the controller nor in the touchscreen, but it could be another type of profile option.

fzurita:
I think another profile type just for accelerometer control would be to much.

If you do it in the touchscreen profiles, you could add a button to toggle it on/off. Also, under that button's configuration in the touch screen profile editor you can configure sensitivity.

It's a good amount of work, but I think that's the right place to put it.

Also, how would a user disable the accelerometer if they don't have a controller? If you put it in the controller mappings like you have it right now they would need a controller to enable/disable it?

fzurita:
Look at how currently the joystick in the touchscreen profile editor handles the animate option. You could do it like that.

javaxubuntu:
I've developed the sensor button in the touchscreen. For the implementation I created a SensorController that now extends AbstractController.
I still have to develop the axis configuration.
What about an option to enable sensor every time? It would be useful in some games.

fzurita:
Sure, additional options don't hurt. You could place the option in the touch screen profile button configuration for the new button you added. You could have default state of either on or off based on the configuration.

javaxubuntu:
or  global option?

fzurita:
Sure, we can try a global option and see how it goes. Thanks for the work on this.

javaxubuntu:
As it may be enabled by default in the touchscreen profile, I'll put the sensor configuration in the "Choose an option" menu of the touchscreen profile configuration:
new menu entry : Choose an option => Sensor configuration

fzurita:
Ok

javaxubuntu:
The feature is implemented:
- In touchscreen profiles, a new menu entry "Sensor configuration" to fully customize X and Y axis, allowing custom values in case we want to play in reverse landscape or portrait mode. The sensor can be activated on start or on demand.
- A new sensor toggle button
- In analog touch configuration: a new checkbox to hide it when sensor is enabled
- In controller profiles, a new button "Toggle sensor".

About the changes:
I added sensor_icon.odg only as the source of the sensor button, it is not used by the app.
PromptInputCodeDialog.java: this is only code refactoring. You can ignore this change if you don't like it.
I also fixed a bug in the "animate joystick" checkbox: its state was not loaded from configuration.

fzurita:
Cool, I'll check this out some time tonight

fzurita:
Ok, this is looking pretty good. A few things that I see. The settings dialog that you added has a strange white border. I'll attach a screen shot. Also, we may want to consider adding automatic flipping and reversing of axis depending on what configuration the user has for screen rotation. For example:
* Landscape: Everything stays as is
* Reverse landscape: Invert Y axis
* Portrait: switch X and Y axis
* Reverse portrait: switch X and Y axis, X is inverted.
screenshot_2015-12-31-00-04-57

fzurita:
I'm going to merge it as is since it's functional enough. Also, I'm going to modify your commit history to make it more streamlined.

I'll keep the pull request open so that you can keep making additional fixes.

fzurita:
Look at this commit: 02e28bbe432c6d82a590b34e30b931b3a16ef124

fzurita:
The code has been merged. You will have to re-pull master to get back in sync since I heavily consolidated your commit history.

javaxubuntu:
"Custom" configuration option allows you to choose any sensor's axis to be used for sensor configuration. Y/xz means that it uses gravity ox Y axis for Strength, comparing this value with the total gravity of X and Z. X/Z means that value does not depend on sensor's Y value.
I think that in portrait mode the values should be X/YZ and Y/Z. (this is case insensitive).

Border was used to group options of the same axis.

fzurita:
Ok, in our UI style, we generally don't use borders like that.

fzurita:
There are a couple of spelling errors. Usual in your settings dialog should be "use".

Also, I don't think there is a need to specify " in landscape mode"

fzurita:
Another thing, custom is not very obvious on how to use. If it's not user friendly enough, I don't think we should present the option.

fzurita:
@javaxubuntu , you may have to do this in your branch from the command line:
"git reset mupen64plus-ae/master --hard"

This will revert your branch to the current status of the master branch here. Before you back up any changes first though since that command will remove any changes that are not in the mupen64plus-ae/master branch.

javaxubuntu:
I've made the axis fully configurable to allow some tests.
Fox example, the Y axis can be calculated with X sensor compared with the total other acceleration
screenshot_2015-12-31-15-16-21
Then when I try it, it's impossible to have the joystick left-down when my phone is vertical turning left with 45°
screenshot_2015-12-31-15-15-30
This is why I chose to calculate Y axis differently than X axis.
Do you want to do some tests with it, or do you prefer I remove the "Custom" axis configuration?

javaxubuntu:
@fzurita 
$ git reset mupen64plus-ae/master --hard
fatal: ambiguous argument 'mupen64plus-ae/master': unknown revision or path not in the working tree.
$ git reset origin/master --hard
HEAD is now at fd8fb1f merge

fzurita:
Ok, cool, I'm glad that command worked for you. About the custom option, it sounds like a useful feature for debugging, but it may not be very useful for the end user. Maybe "custom" should be renamed to "Advanced."

Also, the "Idle Angle" seems to take care of the case when you want your phone to be vertical and be able to turn left and right. Maybe we should consider the vertical position to be the default position?

javaxubuntu:
The Idle Angle is 0 for X axis. This is not useful to have a different angle.
For Y axis, I want the Y axis to be IDLE when my phone is at 45°, not horizontal, not vertical.
screenshot_2015-12-31-16-11-08

fzurita:
Right, I understand.

javaxubuntu:
The sensor config should be written like "global settings"? v7/preference or v14/preference?

fzurita:
Just for the future, I think we need to add the option for the user to hold the phone at a desired position and make that the "idle angle". It could be called " calibrate "

Also, we may also want to consider adding a dead zone.

fzurita:
Everything we do is v7 preferences. Also, we may want to store the preference in each touchscreen profile. Since, I can see people having different accelerometer configurations per game.

fzurita:
No more had been added here for a while. Any additional changes will need a new pull request

PR #531, opened 2015-12-24
Allow mapped controllers to automatically reconnect.

nicklauslittle:
This commit will allow mapped controllers to be remembered by their device "descriptor" rather than the device id. This will allow controllers that have been disconnected and reconnected to be automatically remapped when one of the buttons is pressed. Also, deserialization should be compatible with current saved controller settings.

Note that in-game controller processing still uses the device ids for performance reasons. Also note that pre-Jelly Bean devices must still use device ids, and I was unable to test this with my setup.

### DIFF
--- a/src/paulscode/android/mupen64plusae/game/GameActivity.java
+++ b/src/paulscode/android/mupen64plusae/game/GameActivity.java
@@ -706,7 +706,6 @@ OnPromptFinishedListener, OnSaveLoadListener, GameSurfaceCreatedListener, OnExit
         boolean keyDown = event.getAction() == KeyEvent.ACTION_DOWN;
         
         // Attempt to reconnect any disconnected devices
-        // TODO: Show message on screen when reconnection was successful?
         mGamePrefs.playerMap.reconnectDevice( AbstractProvider.getHardwareId( event ) );
         
         if( keyDown && keyCode == KeyEvent.KEYCODE_MENU )
--- a/src/paulscode/android/mupen64plusae/game/GameActivity.java
+++ b/src/paulscode/android/mupen64plusae/game/GameActivity.java
@@ -705,6 +705,10 @@ OnPromptFinishedListener, OnSaveLoadListener, GameSurfaceCreatedListener, OnExit
     {
         boolean keyDown = event.getAction() == KeyEvent.ACTION_DOWN;
         
+        // Attempt to reconnect any disconnected devices
+        // TODO: Show message on screen when reconnection was successful?
+        mGamePrefs.playerMap.reconnectDevice( AbstractProvider.getHardwareId( event ) );
+        
         if( keyDown && keyCode == KeyEvent.KEYCODE_MENU )
         {
             if( mDrawerLayout.isDrawerOpen( GravityCompat.START ) )
--- a/src/paulscode/android/mupen64plusae/input/map/PlayerMap.java
+++ b/src/paulscode/android/mupen64plusae/input/map/PlayerMap.java
@@ -39,6 +39,20 @@ public class PlayerMap extends SerializableMap
      *  this map allows the old id to be looked up so it can be replaced with the new id. */
     private HashMap<String, Integer> deviceNameToId = new HashMap<String, Integer>();
     
+    public static boolean isDeviceId( String deviceName )
+    {
+        try
+        {
+            if( Integer.parseInt( deviceName ) != 0 )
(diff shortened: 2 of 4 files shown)


src/paulscode/android/mupen64plusae/game/GameActivity.java
```diff
@@ -705,6 +705,10 @@ public boolean onKey( View view, int keyCode, KeyEvent event )
     {
         boolean keyDown = event.getAction() == KeyEvent.ACTION_DOWN;
         
+        // Attempt to reconnect any disconnected devices
+        // TODO: Show message on screen when reconnection was successful?
```
fzurita: Something like this should be invisible to the user in my opinion.

src/paulscode/android/mupen64plusae/input/provider/AbstractProvider.java
```diff
@@ -206,6 +244,34 @@ public static String getHardwareName( int id )
...
+     * 
+     * @return The unique name of the device, or null if hardware not found.
+     */
+
+    @SuppressLint("NewApi")
+	public static String getUniqueName( int id )
+    {
+        if( id > HARDWARE_ID_MOGA_OFFSET && id <= HARDWARE_ID_MOGA_MAX )
+        {
+            return mogaPrefix + ( id - HARDWARE_ID_MOGA_OFFSET );
```
fzurita: This seems correct to me, but I'm unable to test to make sure this works correctly with Moga devices. @littleguy77 Do we have any testers that have moga controllers.

Worst comes to worst, if it doesn't work right, it should still behave like before, they would have to manually re-map the moga controller again to a player.

src/paulscode/android/mupen64plusae/input/map/PlayerMap.java
```diff
@@ -79,7 +88,9 @@ public String getDeviceSummary( Context context, int player )
             if( mMap.valueAt( i ) == player )
             {
                 int deviceId = mMap.keyAt( i );
-                String name = AbstractProvider.getHardwareName( deviceId );
+                String name = AbstractProvider.isHardwareAvailable( deviceId ) ?
+                        AbstractProvider.getHardwareName( deviceId ) :
+                        "Not Connected";
```
fzurita: This should come from strings.xml.

fzurita: I can see why I have more than 1 controller showing up. It's the old saved controller that used to go by device id instead of device descriptor.

To fix this, all you have to do is modify preferences_game.xml. Rename the key "playerMap" to "playerMap1".

nicklauslittle: This sounds like a bug to me. My intention was to deserialize the old ids if the device was still connected, and then serialize it using the device descriptor, without the old id in there. I will review the code and see if I can find anything that would cause this.

fzurita: Here is how you can reproduce it. Install the latest auto build and map a couple of controllers to players. Then install your version of the build and do the same again. You will now have 2 controllers for each player with 1 "not connected".

src/paulscode/android/mupen64plusae/input/provider/AbstractProvider.java
nicklauslittle: Just FYI, I did test with a Moga Power Pro in A mode to make sure I hadn't completely messed something up. Somebody should probably still test it though.

src/paulscode/android/mupen64plusae/input/map/PlayerMap.java
```diff
@@ -125,15 +187,89 @@ public void removeUnavailableMappings()
...
+                    try
+                    {
+                        int player = Integer.parseInt( elements[0] );
+                        int id = AbstractProvider.getHardwareId( elements[1] );
+                        
+                        // If the specified device is not currently connected.
+                        if( id == 0 )
+                        {
+                            // Do not add disconnected devices if the key specified is a device id.
+                            if( Integer.getInteger( elements[1], 0 ) != 0 )
```
nicklauslittle: This (and all other references to Integer.getInteger) should be Integer.parseInt. Integer.getInteger does not do what I though it did. I think this may be the source of all the old device ids showing up even though they are disconnected. I will fix this.

PR #534, opened 2015-12-31
Removed restrictions on buttons and axes mapping to same control.

nicklauslittle:
This commit allows buttons and axes to map to the same controls.

### DIFF
--- a/src/paulscode/android/mupen64plusae/input/InputEntry.java
+++ b/src/paulscode/android/mupen64plusae/input/InputEntry.java
@@ -1,0 +1,82 @@
+/**
+ * Mupen64PlusAE, an N64 emulator for the Android platform
+ * 
+ * Copyright (C) 2013 Paul Lamb
+ * 
+ * This file is part of Mupen64PlusAE.
+ * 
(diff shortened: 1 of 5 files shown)


fzurita:
This works pretty well except for one minor thing. When mapping controls in the controller profile editor when in "Standard mode", normally when pressing keys, the corresponding keys in the virtual controller will become highlighted.

With this change, only the first assigned key makes the corresponding virtual controller key light up, the 2nd assign key doesn't. If you fix that, I think this can go in.

nicklauslittle:
This should be ready. Let me know if you find any more issues. (I wasn't sure if this is still waiting on testing or something. I just thought I would let you that I'm done with it in case it was waiting on me.)

fzurita:
I hadn't realized that you had fixed it. Unfortunately, there are no notifications when additional commits are made to a pull request. 

I'll check this out tomorrow night since tonight I ran out of time.

nicklauslittle:
Yeah, sorry. I should have put in a comment. Next time I'll put one in after I commit.

fzurita:
OK, looks good, all set.

PR #535, opened 2015-12-31
Added global settings for controller profile and player map.

nicklauslittle:
This pulls controller settings into the global settings area. I think this will make the controller setup more  intuitive. Per game settings can still override these settings, but now we won't have to set them up for each game every time we switch games. To revert the game settings to use the global settings, simply unmap all the players and specify the controllers settings as default. I don't think this is the best final solution, but I think it will work better than what is currently there until a better solution is found. Let me know of any issues.

### DIFF
--- a/res/xml/preferences_global.xml
+++ b/res/xml/preferences_global.xml
@@ -212,6 +212,41 @@
             android:key="inputShareController"
             android:summary="@string/inputShareController_summary"
             android:title="@string/inputShareController_title" />
+
+        <paulscode.android.mupen64plusae.preference.ProfilePreference
+            android:icon="@drawable/ic_gamepad"
+            android:key="controllerProfile1"
+            android:summary="@string/selectedValue"
+            android:title="@string/controllerProfile1_title"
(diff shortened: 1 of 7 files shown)


fzurita:
I agree that this needs work in the current implementation. Your change should get us in the right direction. Ideally, I think the right solution will be to have a player setup menu in the settings side bar.

fzurita:
Also, this is currently crashing when assigning a controller to a player under global settings. See this crash log.
[Crash_2016-01-02_14-50-12_893.txt](https://github.com/mupen64plus-ae/mupen64plus-ae/files/76376/Crash_2016-01-02_14-50-12_893.txt)

fzurita:
I can see why it's crashing. Make sure you get in sync with the latest master. It will crash for you there as well. It's crashing because the android:key="playerMap" under preferences_global.xml should be playerMapV2.

I'll go ahead and make the fix for you.

fzurita:
See: e86f3f56820cac06bab20e52823f98d8b10d4fb6

Issue #540, opened 2016-01-12
"Failed to attach core library" on Marshmallow


nicklauslittle:
I get a "Failed to attach core library" on my Nexus Player running Marshmallow on all builds after 12/09. I think this is just a generic marshmallow issue, but I could be wrong.

I did a little bit of digging, and I think I know what the problem is. (No guarantees.) It looks like the mupen64plus-ui-console library is failing to load the mupen64plus core library. I did a quick search and found that on apps that target pre-Marshmallow versions attempting to load a library with text relocations would issue a warning but still succeed. When targeting Marshmallow, this fails with an error. To attempt to verify this I built two apks based off commit 33036d5f. The first targeted version 23 (AndroidManifest.xml:18) of the SDK and issued the error above. The second targeted version 22 and it loaded the core and worked just fine.

I don't have a clear view of how to fix this. My guess would be that it would require modifying some of the core source or some compile options, or both. I do plan on looking into it at some point, but I figured I would post it here in case someone knows what needs to be done.

fzurita:
Right after that we switched everything to SDK version 23. Oddly enough, I tried it on my nexus player running marshmallow and I don't get that error. The difference may be that I'm running a custom ROM that disables leanback.

fzurita:
Can I get your log cat output?

nicklauslittle:
Maybe. I had two items in the "Share" list (from ES File Explorer) and wasn't able to get it to export using either one. I may be able to pull them through adb, but I'm not familiar enough with Android development to easily accomplish this yet. I may try to tackle it this weekend. Let me know if you know of an easier way to do this.

However, I did look at the logs and can confirm that it is what I originally thought. I can run the target 22 build and see the "... has text relocations. This is wasting memory and is a security risk. Please fix." message as a warning. When I load the target 23 version I see the message as an error and more errors follow it indicating it could not load the core.

I found some sites that talk about this. (One is https://code.google.com/p/android/issues/detail?id=67770) I may attempt some of the recommended actions and see if I can get anything to work correctly that targets version 23.

fzurita:
In Android.mk under the jni directory, add ```-fPIC``` to ```COMMON_CFLAGS```.

I got this from the link you posted.

nicklauslittle:
[...]

PR #542, opened 2016-01-23
Fixed text relocation errors on x86 Android M

nicklauslittle:
Builds 2 shared objects for the core on x86 and switches between them at run-time depending on if Android M or later is detected. This fixes https://github.com/mupen64plus-ae/mupen64plus-ae/issues/540. Known Issue: Cannot resume a game when the PIC-compliant version of the core (old dynamic recompiler) is loaded.


fzurita:
Thanks :)

PR #543, opened 2016-01-24
Hack to temporarily fix issues with resuming on x86 Marshmallow.

nicklauslittle:
I discovered that the old dynamic recompiler (for x86 on Marshmallow and newer) apparently needs a little bit of warm-up time before it can load saves. I tested this on my Nexus Player to determine how long of a warm-up was needed. It was somewhere between 250ms and 500ms, so I doubled that to 1000ms. It still may not work every time on everyone's setup, but I haven't been able to get it to fail. Hopefully, this is only a temporary fix until the new dynamic recompiler can support Marshmallow on x86.


fzurita:
Looks good to me

PR #544, opened 2016-02-08
New Video plugin

fzurita:
@Gillou68310 @littleguy77 I ported over the libretro implementation of the Glide64 plugin.

Do you think it makes sense to include this version of the plugin in the master branch? Or should we hope that eventually the upstream Glide64 will eventually get all the changes of the libretro Glide64?

Also, ignore the branch name. It was convenient at the time to do the work for this plugin in the "busywait" branch since it contained a fix I wanted. I probably should had made a new branch off the busywait branch.


Gillou68310:
> Do you think it makes sense to include this version of the plugin in the master branch?

What is the performance + accuracy compared to Glide64mk2 plugin?

Gillou68310:
> @Gillou68310 does this change make sense to you? I'm having retroben test it when he gets a chance, but I have high hopes for it.
> 
> Do you think upstream would take this? The change is basically, if we sleep for too long, then sleep less the next time around.

Sorry sounds like I missed your this one :-(
I saw you already opened the PR upstream, let's wait for richard42 analysis ;-)

fzurita:
No problem on missing that comment.

On regards to the glide2gl plugin (what libretro calls it). I haven't tested performance and accuracy personally. I also still haven't made parameters adjustable to make a fair comparison in performance.

The only reason I did this was because I saw that it has been continuously updated way more than the original glide64mk2 plugin has been and because retroben was saying that it was faster compared to glide64mk2.

Gillou68310:
IMO we should avoid adding another video plugin, if glide2gl is proved to be better than glide64mk2, I guess we'll have to remove the latter.

fzurita:
It definitely seems more stream lined. Also, I may be wrong, but it seems like the Glide wrapper is mostly gone which does remove some of the bloat.

Gillou68310:
I still see Glitch64 glide wrapper in the code, so I guess it's not gone ;-)

fzurita:
I didn't look too carefully, but it seems like some of the Glide specific calls were removed. At least from the "main" file.

fzurita:
If we were to get rid of the glide64mk2 video plugin, we would lose the convenience of being able to pull directly from upstream into our build. Also, each time that libretro updates glide2gl, we would have to cherry pick the commits due to big differences in their baseline.

PR #545, opened 2016-02-24
Glide2gl

fzurita:
@Gillou68310 @littleguy77 Please see original pull request #544. Now this is referencing the correct branch.


fzurita:
I'll get rid of this pull request, no good reason to use it I think. There haven't been any important updates from them.

PR #589, opened 2016-10-05
Lithuanian translation added

zygimantus:


fzurita:
Thanks! I'll integrate it.

fzurita:
Is this intended to only translate a few strings?

zygimantus:
No, it is just the beginning. I will try to translate everything.

fzurita:
I'm going to go and merge this as is, since some translation is better than nothing.

PR #614, opened 2017-03-07
Update SDL2

loganmc10:
This is definitely not done, I just wanted to create this so you could see what I'm working on and provide input.

This commit updates SDL to 2.0.5, I'm not actually sure what the current version of SDL2 being used it, but it looks like it's from around 2013.

The SDL2 folder was simply updated using the SDL2 sources, no manual changes inside that folder.

The biggest changes involve the EGL context creation. In the new version of SDL, you don't have direct access to context creation or buffer swapping (no more createGLContext, deleteGLContext, or flipBuffers). The context creation and buffer swapping is handled by SDL.

The biggest issues remaining:

- SDL_EGL_SwapBuffers is being called when it shouldn't (crashes when resuming the app, and sometimes on close). The current implementation checks if the context is OK before calling flipBuffers, since there is no flipBuffers anymore, I'm not sure what to do about this yet.
- I haven't done extensive testing, I'm not sure if full OpenGL context creations still work for GLideN64. This might already be a non issue.
- The FPS counter is broken, since that happened inside the "flipBuffers" function, again I'm not really sure how to solve this since you don't have access to the flipBuffers function anymore

The only real way to have more control over the SwapBuffers function is to implement Mupen64Plus's Video Extension API. The docs are here:
https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Video-Extension.mediawiki

Basically, you tell mupen64plus that you don't want SDL to handle the video, you then need to implement the various GL interface functions like VidExt_SetVideoMode, VidExt_GL_SwapBuffers, etc...

You would then be able to do whatever you want inside VidExt_GL_SwapBuffers, like checking if it's safe to swap buffers, implement the frame counter, etc...

The problem with this approach is that you need to call CoreOverrideVidExt() after the core library is initialized, and before the emulator is running, since mupen64plus-ae interfaces with mupen64plus-ui-console and not directly with the API to kick things off, that may be challenging.


loganmc10:
I haven't looked at the code too closely, but the SwapBuffers issue is more an indication of a larger problem. It shouldn't be swapping buffers when the screen isn't visible, or when the emulator is paused. So it seems like the emulator is running when it shouldn't be.

The pause/stop commands execute asynchronously, so it shouldn't be assumed that they happen as soon as the command is run. You have to do a ```CoreDoCommand(M64CMD_CORE_STATE_QUERY, M64CORE_EMU_STATE, &Value)``` and check the state of the emulator to be sure if it's paused or stopped.

I don't know too much about how Android handles app closing, but if there is some way to catch that signal, and run a loop that waits until M64CMD_CORE_STATE_QUERY returns "paused" or "stopped", then that may avoid SwapBuffers calls at the wrong time.

fzurita:
This is cool. About the FPS, we are doing it wrong right now. Mupen64plus provides an API exactly for this that we are not using for some reason.

Another thing... We should not be using UI console in the first place. The UI itself should be doing this. The ui-console does have some cheat related code that we just need to take out and copy instead.

fzurita:
The check for null on the swap buffers call I don't think is needed any more. I think that was before I made some fixes to an issue that was causing the core to still be active after the activity was gone.

fzurita:
That crash that you are experiencing may be related to resuming the core before the GL context has been created. I'm not sure how the new SDL manages recreating the context once the activity has resumed. It doesn't seem to be doing it early enough. Maybe there is a way to tell it to recreate the context sooner.

loganmc10:
> Another thing... We should not be using UI console in the first place. The UI itself should be doing this. The ui-console does have some cheat related code that we just need to take out and copy instead.

I chuckled when I saw that's how it works, it's actually not a bad solution, given how simple ui-console is, you just lose some flexibility, and I assume it increases the size of the apk a bit.

fzurita:
Right, it's an unnecessary library to have, although it is small. It was probably done like that at the beginning out of laziness.

loganmc10:
I'm still having some trouble getting things to happen properly on Pause/Resume of the app, I'll play with it a bit over the next few days.

I guess the proper way to do the FPS counting would be to use M64CMD_SET_FRAME_CALLBACK and count the FPS in the callback function? I don't see any other way to do it.

fzurita:
Yeah, that would be the correct way of doing it.

fzurita:
I just thought about this. Because this old version of SDL allows the app to create its own OpenGL context, some nasty hacks can be used to give mupen64plus Vulkan support. For example, if we wanted to bring in ParaLLEl, we would have to resort to that. With the new SDL version though, that becomes a lot more difficult to do.

loganmc10:
Ya having SDL do it is kind of a pain, I'm having a hard time getting it to work right anyway.

It would be better to implement mupen64plus's VidExt API and use EGL to handle everything, then you have really good control over the whole system.

In order for that to happen though, you can't use ui-console anymore. You have to call CoreOverrideVidExt after the core library is initialized and before the emulator is running, ui-console doesn't do that.

The easiest way to do that would be to just rename mupen64plus-ui-console to mupen64plus-ui-interface (so it's clear it's not a 1:1 copy of ui-console) or something and just add the CoreOverrideVidExt line in the main() function.

There are of course other ways to remove the dependency on ui-console, like re-write it in Java or something, it's up to you how that gets accomplished. Once the dependency on ui-console is removed, I don't mind writing the VidExt API stuff, I recently did that for a Qt frontend I wrote (https://github.com/m64p/mupen64plus-gui/blob/master/vidext.cpp), so I have a pretty good understanding how all that is supposed to work.

If the VidExt API gets implemented, then updating SDL2 is no problem, since all of the SDL video stuff gets bypassed.

fzurita:
Cool, I wasn't aware of the VidExt API. Then it would make it a non issue.

loganmc10:
Ok well I'll come back to updating SDL2 another day, I think first it would be wise to move away from having SDL manage the GL context.

I've managed to get the VidExt thing working, see here:

https://github.com/loganmc10/mupen64plus-ae/tree/vidext

Basically there is now an "ae_vidext" library, written in C, that handles everything to do with EGL (context creation, buffer swapping, etc...). It can also be integrated into the existing mupen64plus-ui-console with a small change upstream (using the same method as the CALLBACK_HEADER function).

I still need to fix the context creation code (right now it always creates a GLES2 context), and I'm sure it doesn't work yet with pausing/resuming, but the basic building blocks are there and it seems to work well.

fzurita:
Cool. That's actually really useful. It would allow us to update to the new SDL without having SDL manage the context creation.

loganmc10:
I assumed that performance (mostly of the swapbuffers call) would be better if it was written in C. I wasn't sure what the overhead of a C->Java (and then probably back to C via Android's EGL14 library) trip would be like, but maybe I'm overthinking it.

fzurita:
That does bring a good point on the swapbuffers call. One call ever 16 to 32 milliseconds (60 or 30fps) shouldn't be too bad, but the only way to tell for sure would be to measure how long it takes to make the call. I suspect it would be negligible compared the rest of the time spent on other places in the code, like 10s of microseconds to make the call.

Only way to tell for sure though would be to measure it.

fzurita:
Googling for some actual numbers doesn't seem to be easy. Best I could find was this from 5 years ago:
http://stackoverflow.com/questions/13973035/what-is-the-quantitative-overhead-of-making-a-jni-call

So I guess I'll have to add some logging to determine how long it takes. Calling the C++ code directly will definitely be faster, I guess the question is how much.

Also, the JNI does add an unneeded level of complexity that is probably not needed. Really, the only reason to keep the code in the java side would be to allow re-use of the existing initialization procedure. So if you really are up to re-writing that portion, I see no problem in doing it all in C/C++.

loganmc10:
So the VidExt thing is going really well, with one major problem. The gln64 plugin violates the API and calls SDL directly, it doesn't go through the VidExt API, see this for example:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/app/src/main/jni/mupen64plus-video-gln64/src/OpenGL.cpp#L317
https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/app/src/main/jni/mupen64plus-video-gln64/src/OpenGL.cpp#L235

It doesn't call any of the VidExt API functions. So it will need to be re-written to go through VidExt like the other plugins, and not call SDL directly.

fzurita:
That's not a problem. glN64 for mupen64plus-ae is a forked version. If you get it working with the other plugins, I can take care of glN64.

loganmc10:
Ok, it's actually probably not a lot of work, the functions are listed here:

https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Video-Extension.mediawiki

Basically instead of doing ```SDL_InitSubSystem``` you do ```VidExt_Init```. Instead of ```SDL_GL_SwapBuffers```, you do ```VidExt_GL_SwapBuffers```, etc...

fzurita:
To be truthful, we really should just delete glN64. It's basically GLideN64 without frame buffer emulation and much less accurate.... It's the worst of all the plugins we include.

fzurita:
Cool, if you feel like updating glN64, feel free :).

loganmc10:
I opened a PR for the VidExt stuff (https://github.com/mupen64plus-ae/mupen64plus-ae/pull/615), I'll close this and re-create a new PR for SDL2 once it can just be dropped in

fzurita:
Ok

PR #615, opened 2017-03-10
VidExt API Implementation [WIP]

loganmc10:
This is getting much closer to being complete. GLideN64 and Rice Video should work perfectly now, although I still need to test GLideN64's full GL profile on my Shield Tablet (I've been testing on my phone).

TODO:


fzurita:
Yeah, SDL2 is really not needed. It has always been worse quality and adds significant audio lag. I wouldn't mind removing it.

fzurita:
I just tried my Adreno device since you were worried about those. It seems to work just fine. I would be more worried about different Android versions since not every EGL configuration can be supported by every device.

loganmc10:
What is the minimum version of EGL supported? They all support EGL 1.4 now don't they since you dropped support for older versions of Android? That should be sufficient

fzurita:
Yep, the project version now requires Android 4.4 which is guaranteed to have EGL 1.4. So everything is probably fine. Did you end up testing full GL support?

loganmc10:
No not yet, I've been juggling a few things and haven't gotten around to that yet. It should work fine though. The way it works now t properly interacts with the mupen64plus API, so it respects the fact that GLideN64 requests an OpenGL 3.3 core profile.

I have it fallback to a GLES2 profile, so basically with GLideN64 it will request a 3.3 core profile, that will fail if it's not supported, then it will create a GLES2 profile, which gets upgraded to GLES3/3.1/3.2 if that is supported, since they are backwards compatible, but I still need to test it to make sure that all works on the Shield Tablet.

loganmc10:
Alright, all I need to do is do some testing on my Shield Tablet, but this should be complete, in summary:

- Dependency on SDL Video has been removed, all of the EGL initialization and buffer swapping has been moved to ae-vidext. This should make it fairly simple to implement a Vulkan backend as some point as well.
- Audio SDL has also been removed, as well as all the SDL->Java tie ins. The newest version of SDL2 can simply be dropped in the JNI folder after this is merged.
- gln64 has been refactored to use the VidExt API instead of using SDL directly, so all the video plugins should work the same as before.
- The FPS counter has been re-implemented inside the ae-vidext "VidExtFuncGLSwapBuf" function.

Hopefully this all doesn't seem too intrusive, I think it's a better way to handle the EGL code, it should offer more flexibility, the documentation on what each VidExt function is supposed to do is found here:

https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Core-Video-Extension.mediawiki

fzurita:
Cool, thanks, I'll take a detailed look tonight.

fzurita:
Along with OpenGL, can you also check that you are able to reach above 60fps when fast forwarding? I had to do a hack for that, setting the "presentation time" to a very small value when fast forwarding, so that frames could be skipped.

loganmc10:
Ok I'll look into that, my feeling is that is probably happening because eglSwapInterval defaults to 1, so vsync is on by default, which will cap the FPS at the refresh rate.

With VidExt the video plugin can call VidExt_GL_SetAttribute(M64P_GL_SWAP_CONTROL, 0/1) to enable/disable vsync, which will in turn call eglSwapInterval in this case.

GlideN64 does this, and it defaults vsync to 0, so with this new VidExt API, GLideN64 will have vsync off by default, whereas it was probably on before.

We can decide whether vsync should be on or off by default (but the video plugin will be able to change it). we can also turn it off when fast forwarding

fzurita:
Vsync should be on when not fast forwarding and off otherwise. I didn't think that Android let you turn off vsync though. The only workaround I could think of was using the presentation time. Maybe it's different when one stays in the C++ world.

loganmc10:
Ok so full OpenGL works, it reports the GL version as 3.3 on the Shield Tablet. That is kind of strange, since it supports higher, but that is the exact same thing Nvidia does on the desktop, it reports 3.3, so it's working as expected.

As far a fast forwarding it, it seems like you are right. With Mario 64 it speeds up, but Smash Bros doesn't. I assume since Mario 64 is 30FPS and Smash Bros is 60FPS.

If I understand it right, what I need to do is re-implement this right before swapping buffers?

```
if(NativeExports.emuGetSpeed() > 100) {
    EGLExt.eglPresentationTimeANDROID(mEglDisplay, mEglSurface, System.nanoTime() - 4000);
}
```
the C equivalent is just:
```
eglPresentationTimeANDROID(EGLDisplay dpy, EGLSurface surface, EGLnsecsANDROID time);
```

Just so I make sure I understand, ```System.nanoTime()``` is the current system time in nanoseconds?

loganmc10:
Follow up question mostly out of curiosity. What is it doing exactly? Isn't it specifying a time slightly in the past?

fzurita:
Yes, it is specifying a time in the past. So that any queued frames immediately time out. The eglSwapBuffers call is actually queueing a frame, the presentation time sets a timeout for the queued frame, so that if it can't be displayed within the timeout, it's dropped.

fzurita:
I think overall it looks pretty good. I'll have to make sure some situations still work before merging in. Also, since SDL is gone, we don't need libsamplerate any more.

loganmc10:
Ok I re-implemented the fast forward fix and I tested it, it works.

I was reading the documentation on eglPresentationTimeANDROID, and it is a little unclear to me. Does that call affect all future eglSwapBuffers call? Or just the next one? If it affects all future calls, then it should be reset to default once fast forwarding has stopped, but I can't find anything about what the default should be.

fzurita:
Correct. Here a good overview if you are curious: http://stackoverflow.com/questions/26317132/minimize-android-glsurfaceview-lag/

loganmc10:
That guy says:

> one way to avoid queueing up frames is to call eglSwapInterval(0). If you were sending output directly to a display, the call would disable synchronization with VSYNC, un-capping the application's frame rate. When rendering through SurfaceFlinger, this puts the BufferQueue into "async mode", which causes it to drop frames if they're submitted faster than the system can display them.

That didn't seem to work, eglSwapInterval(0) should be getting called for GLideN64 now, but it didn't seem to allow fast forwarding, anyway, this approach seems to work well.

loganmc10:
I realized what I was doing wrong, I was called eglSwapInterval before the context had been created. I moved it to the proper place, and now fast forwarding works fine when vsync is disabled. I've rebased the commits and removed the eglPresentationTimeANDROID code. I think using eglSwapInterval is probably the better way to do it, but I can revert it if you'd like.

The problem becomes: what to do about vsync. Right now it defaults to 1, but the video plugin can override that setting. GLideN64 does, it defaults to vsync = 0, so fast forward works there, but I don't think the other plugins do anything with that setting.

Should vsync default to 0 instead of 1? Or should vsync get disabled if fast forward is happening and then set back to the plugins desired value afterwards?

fzurita:
Ok, that's good to know. I think we should ignore the plugin setting. Disable vsync when fast forwarding is disabled, enable it otherwise.

fzurita:
What device are you using to test eglSwapInterval? This doesn't seem to be supported by all the drivers from the quick Google search I just did.

loganmc10:
I'm testing this on an Xperia XA (Mali T860). I still need to test on the Nvidia Shield.

fzurita:
Ok

loganmc10:
Ok I've go that implemented here:

https://github.com/mupen64plus-ae/mupen64plus-ae/pull/615/commits/8355e62a6a331cb0f8d799921c6301a65944fab9

vsync is on by default (ignores plugin setting), and off when speed is over 100%. Tested on my Mali phone, I'll test on the Shield in a bit, but I assume if the Mali can do it the Shield can.

fzurita:
Ok, that's good enough of a test for me. I'll test my Adreno device just to make sure. It's probably unreliable in the older Android versions.

loganmc10:
Devices can definitely vary in their support of the values that can be passed to eglSwapInterval. A device can query EGL_MIN_SWAP_INTERVAL and EGL_MAX_SWAP_INTERVAL to get the constraints, but I think every device nowadays should support at least 0 or 1. I do see some posts from ARM/Mali from 2011/2012 where they say their devices don't support eglSwapInterval(0).

Probably the same devices that support eglPresentationTimeANDROID will support eglSwapInterval(0) would be my guess.

loganmc10:
Fast forwarding works on the Shield Tablet as well

fzurita:
I tried fast forwarding in my Adreno device and it works fine as well. I always thought that the way I did it with the presentation time was kind of hacky anyways.

fzurita:
I think this is ready for the most part, any minor issues that show up we can fix in the master branch or submit additional pull requests.

fzurita:
Oh, and thanks by the way.

loganmc10:
My pleasure

PR #648, opened 2017-08-18
Support Google Cardboard VR

justinmichaud:
screenshot_20170817-214125

I just wanted to know, if I would clean this up and put it behind a config option, would you be willing to merge it? It is a big mess right now, but excepting a few small graphical glitches, it seems to work. At the same time, I could understand not wanting to take on the additional maintenance burden. 

If you want to try it out, use the gladen64 plugin + widescreen fix.

Thanks for making such a cool emulator!

PR #649, opened 2017-08-19
Cardboard VR support (rebased onto latest GLideN64)

justinmichaud:
Original PR: https://github.com/mupen64plus-ae/mupen64plus-ae/pull/648
GLideN64 PR: https://github.com/gonetz/GLideN64/pull/1552

This PR is massive because I deleted gliden64 and replaced it with a submodule so that it is easier to track the GLideN64 PR.

Once the GLideN64 PR is merged and if you are fine with using submodules, all that would need to be done is to merge this PR + change the submodule directory to point to the real project instead of my fork. Otherwise, you can cherry-pick the config option commit (581b67e) and then update GLideN64.

Thanks again for the help!


fzurita:
Cool thanks! Once the GLideN64 pull request is accepted, I will cherry pick the config updates.

I'm hesitant with going with sub modules at this point. I would rather wait until it's ready upstream. I will definitely give this a shot though before that.

PR #723, opened 2019-01-31
Add a built-in profile for raphnet-tech GC/N64 to USB V3

raphnet:
The raphnet-tech GC/N64 and N64 to USB adapters do not work
with the built-in profile, or at least not on all systems.

This patch adds a new built-in profile that works out of the box
with raphnet v3 adapters. Tested on a Samsum SM-J320V running android
7.1.1.

### DIFF
--- a/app/src/main/assets/mupen64plus_data/profiles/controller.cfg
+++ b/app/src/main/assets/mupen64plus_data/profiles/controller.cfg
@@ -10,6 +10,12 @@ map=0:201,1:203,2:202,3:200,4:197,5:196,6:190,7:189,8:-30,9:-29,10:-23,11:-24,12
 deadzone=0
 sensitivitiy=100
 
+[USB N64 Adapter (raphnet V3)]
+comment=raphnet-tech GC/N64 and N64 to USB adapters, version 3
+map=19:-4,18:-3,17:-2,16:-1,7:96,6:97,4:99,11:102,10:103,9:104,8:105,0:106,2:108,3:109,1:110,
+deadzone=0
+sensitivity=100
+
 [Xbox]
 comment=Xbox360 and XInput-compatible PC controllers
 map=0:-31,1:-32,2:-33,3:-34,4:108,5:-23,6:99,7:96,8:-25,9:-26,10:-27,11:-28,12:103,13:102,16:-1,17:-2,18:-3,19:-4,32:97,
--- a/app/src/main/java/paulscode/android/mupen64plusae/task/ExtractAssetsTask.java
+++ b/app/src/main/java/paulscode/android/mupen64plusae/task/ExtractAssetsTask.java
@@ -89,7 +89,7 @@ public class ExtractAssetsTask extends AsyncTask<Void, String, List<ExtractAsset
             mAssetVersions.put("mupen64plus_data/gln64rom.conf", 1);
             mAssetVersions.put("mupen64plus_data/mupen64plus.ini", 4);
             mAssetVersions.put("mupen64plus_data/mupencheat.default", 1);
-            mAssetVersions.put("mupen64plus_data/profiles/controller.cfg", 1);
+            mAssetVersions.put("mupen64plus_data/profiles/controller.cfg", 2);
             mAssetVersions.put("mupen64plus_data/profiles/emulation.cfg", 3);
             mAssetVersions.put("mupen64plus_data/profiles/touchscreen.cfg", 2);
             mAssetVersions.put("mupen64plus_data/skins/touchscreen/JoshaGibs/analog-back.png", 1);


raphnet:
I would like to add that the existing built-in profile [USB N64 Adapter] does not work with my old raphnet.net V2 adapter. I'm beginning to wonder if the buttons numbers are constant across devices and/or Android releases...

For reference, here is what I get when configuring a V2 raphnet adapter:

map=19:-4,18:-3,17:-2,16:-1,7:96,6:97,5:98,4:99,11:100,10:101,9:102,8:103,13:104,12:105,0:106,2:108,3:109,1:110,

fzurita:
You are right. Some devices handle USB controller different depending on what controller drivers are installed or how the drivers are configured in the Android system. The best one can do is have a profile that works for most devices.

Can you update your pull request to bump the controller.cfg version number here to 2?

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/app/src/main/java/paulscode/android/mupen64plusae/task/ExtractAssetsTask.java#L92

fzurita:
Yep, that's understandable. We will see based on how many people say it works or doesn't work.

fzurita:
Oh, and thanks :-)

fzurita:
In the end, I think the correct way to handle raphnet devices would be through Android USB host mode, which basically allows raw USB device access through an Android API.

The API is documented here: https://developer.android.com/guide/topics/connectivity/usb/accessory

I was intending to do it at some point by taking your implementation (https://github.com/raphnet/mupen64plus-input-raphnetraw) and switching it to the Android API, but I just haven't had the time.

raphnet:
Yeah, that would be nice. I'd like to look into this. Perhaps I could take care of the low level details and make a simple class which you could then use in the emulator?

Otherwise, about the the current issue, I just created this guide to help users do manual controller configuration: https://www.raphnet-tech.com/support/mupen64plus-fz/index.php

fzurita:
If you were to make such, utility class, I would definitely incorporate it. Although the bad thing about that API is that there doesn't seem to be a way to use it under the C++ environment (NDK) of Android. I did read somewhere though, you one could pass the file handle generated by the API into the C++ code and simply use it like a file.

fzurita:
https://www.raphnet-tech.com/products/n64_usb_adapter_gen3/index.php

Is this the right controller adapter to buy? I can take care of all the high level stuff like requesting access to the USB device and retrieving a file handle for it, then passing it to a C++ class, if you are willing to implement the low level commands.

raphnet:
Yes, that's the right one. The 2 player version or the Gamecube/N64 version are also usable.
https://www.raphnet-tech.com/products/dual_n64_to_usb_adapter/index.php
https://www.raphnet-tech.com/products/gcn64_usb_adapter_gen3/index.php

I will have an nvidia shield soon for development and testing. I'll try to start working on this next week. UsbDeviceConnection indeed has a getFileDescriptor method which looks promising.

fzurita:
Ok, I ordered the two player version. Now... to find a N64 controller. Any recommendations for a good source?

fzurita:
Actually, I ended up ordering this one: https://www.raphnet-tech.com/products/dual_n64_to_usb_adapter/index.php

Was that a mistake?

raphnet:
No it's perfect. It will be shipped today.

As for a good source for N64 controllers, I would recommend the "Hard Off" stores in Japan, but this probably does not help unless you are there. Other than that I don't really have one.. so eBay I guess?

The retro fighters controller is a recent 3rd party N64 controller that works fine, but it's 3rd party..

fzurita:
Ok, ebay it is then.

fzurita:
Do most 3rd party controllers work ok with the mempak and rumble pak?

fzurita:
Ok, got some 3rd party rumble paks, memory cards, and two authentic N64 controllers with ok sticks according to the seller descriptions.

fzurita:
Ok, I received two controllers and the adapter. I'll get started on the Java portion tonight. I don't think it will take too long to get that part working.

fzurita:
Here is what I see for the device in Android. To be truthful, I'm not sure which device or endpoint I should grab a file handle for:
```
    UsbConfiguration[mId=1,mName=null,mAttributes=128,mMaxPower=25,mInterfaces=[
    UsbInterface[mId=0,mAlternateSetting=0,mName=null,mClass=3,mSubclass=0,mProtocol=0,mEndpoints=[
    UsbEndpoint[mAddress=129,mAttributes=3,mMaxPacketSize=16,mInterval=1]]
    UsbInterface[mId=1,mAlternateSetting=0,mName=null,mClass=3,mSubclass=0,mProtocol=0,mEndpoints=[
    UsbEndpoint[mAddress=130,mAttributes=3,mMaxPacketSize=16,mInterval=1]]
    UsbInterface[mId=2,mAlternateSetting=0,mName=null,mClass=3,mSubclass=0,mProtocol=0,mEndpoints=[
    UsbEndpoint[mAddress=131,mAttributes=3,mMaxPacketSize=64,mInterval=1]]]]
	
```

fzurita:
Ok, initial cut at the java code has been made. It can be found here:
https://github.com/mupen64plus-ae/mupen64plus-ae/commits/raphnet_adapter

I grab raw access to a device with a vendor ID of 10395 and product id of 100, which is what the adapter that I received showed. I then pass the file descriptor for the USB device to the C++ code.

You can tweak the vendor and product IDs here:
https://github.com/mupen64plus-ae/mupen64plus-ae/blob/79229433b352af2419966f44ceb6c80d75e94446/app/src/main/java/paulscode/android/mupen64plusae/jni/RaphnetControllerHandler.java#L56-L57

I imagine that in the end I will have a list of them. Also, the USB device I received has 3 USB interfaces. The selected interface can be tweaked by modifying ```RAPHNET_INTERFACE``` in the same above class.

I pass in a file descriptor here to the C/C++ code:
https://github.com/mupen64plus-ae/mupen64plus-ae/blob/79229433b352af2419966f44ceb6c80d75e94446/mupen64plus-input-raphnet/upstream/src/plugin.c#L58

The file can be turned into a c++ project by renaming it to plugin.cpp and changing the Android.mk file appropriately.

And finally, to enable raw USB access, once you start the app, go to settings, the input, and enable Raphnet Raw USB access there.

fzurita:
I went ahead and made the plugin.cpp into a C++ file.

fzurita:
And I stand corrected, it seems that hidapi can be compiled against Android: https://github.com/signal11/hidapi/blob/master/android/jni/Android.mk

raphnet:
Yes, hidapi is available for Android and uses libusb. I don't see an obvious way to bypass enumeration and work with a device directly using a file descriptor though. So either I'll make the required changes to hidapi/libusb or make something custom by taking bits from both.

fzurita:
I see this:
```
Users have reported success in using android.hardware.usb.UsbManager
to request permission to use the UsbDevice and then opening the
device. The difficulties in this method is that there is no guarantee
that it will continue to work in the future Android versions, it
requires invoking Java APIs and running code to match each
android.hardware.usb.UsbDevice to a libusb_device.
```

So, you may not need a file descriptor at all, just make sure you only use the device that the java side has requested access to.

fzurita:
I'll try to get hidapi and libusb incorporated into the app.

raphnet:
Thanks! That's actually the hard part for me (figuring out how to deal with native libs in Android Studio)

raphnet:
If just opening the same device using hidapi/libusb work as suggested, we're almost there

fzurita:
Ok, I updated the raphnet_adapter branch. It now incorporates libusb and libhdmi. Both libraries are compiling successfully.

raphnet:
Thanks, that was fast! I just finished installing the Nvidia shield and have it setup for debugging over the network. I'll build the raphnet_adapter branch now.

It may take me a few days before I have something as I'm working of a lot of things simultaneously at the moment...

fzurita:
Yeah, no problem. I also incorporated your input plugin. It's actually building and deploying just fine under Android. It's probably not working though.

fzurita:
The upstream libusb doesn't seem to support opening file handles passed to it. I instead found a fork of libusb that does allow that using this call:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/raphnet_adapter/mupen64plus-input-raphnet/libusb/libusb/core.c#L1249

raphnet:
Thanks. This afternoon I played a bit with the upstream version too and indeed enumerating devices did not work. I have already started modifying hidapi and looked at a few forks of libusb for Android, but the more I think about it the more it all seems overkill... I'm very tempted to just update my plugin to directly do ioctls on the file descriptor instead of going through two layers of modified libraries.

Still, i'll experiment with the fork tomorrow to see how it goes.

raphnet:
I have started updating the pluging. There will be a gcn64_android.c file better suited for this context. Ideally I would need to receive a list of file descriptors to support multiple adapters. Also, could Java  also forward the VID/PID of the device associated with each descriptor? The plugin needs this info to use the correct transfer size and to know how many ports the adapter has.

fzurita:
Yeah, java can forward those things. I will add to the native interface. The easiest thing for me to do would be to provide a separate call for ever single Raphnet device that has been detected. I will work on these things next, it shouldn't take too long.

fzurita:
Also, I found the list of VIDs and PIDs in your code. I can make java look for all of those as well.

raphnet:
If you're going to be passing the VID/PIDs, I would just have the Java code match the VID (0x289b). Then when I release a firmware with a new PID, there would be only one place to update.

The table in my code also has information about the interface to claim. I will try claiming the correct interface from the native side.

fzurita:
ok

fzurita:
The branch has been updated with the above.

raphnet:
Ok, I got it working! I'm using the libusb fork you included directly from my plugin code, so hidapi is no longer required.

fzurita:
Nice! That was fast

raphnet:
But it needs a lot of cleaning up now. In a few hours I will push something usable.

fzurita:
Ok

raphnet:
Ok, here is the first working version: https://github.com/mupen64plus-ae/mupen64plus-ae/pull/726

PR #764, opened 2019-11-10
core: Update to commit eb5ffc4.

bladeoner:
mupen64plus/mupen64plus-core@eb5ffc4


fzurita:
Thanks!

Not much has changed in the core for a while, so I hadn't bothered to update it.

PR #771, opened 2019-11-14
rsp-cxd4: Update to commit 7b99972.

bladeoner:
mupen64plus/mupen64plus-rsp-cxd4@7b99972


fzurita:
Last I checked, this latest CXD4 rsp is causing issues. So I don't think wet should update at this time until upstream issues are fixed.

bladeoner:
Hi I'm figuring out what is going on.

bladeoner:
@fzurita I will close this for now and look into https://github.com/cxd4/rsp later, if we can backport some things.

PR #779, opened 2019-11-24
video-angrylion-plus: Update to commit 7980924.

bladeoner:
https://github.com/ata4/angrylion-rdp-plus/commit/7980924


fzurita:
Hmmm, this updates the source code. But angrylion is no longer being compiled?

bladeoner:
@fzurita I just re-added the files, I ran into some trouble building this commit.

When you refresh the PR, I merged the commits

bladeoner:
I'm still investigating what is going wrong.

fzurita:
Upstream they got rid of the Android makefile. This means that wet have to use CMake through Gradle to make it work.

bladeoner:
> Upstream they got rid of the Android makefile. This means that wet have to use CMake through Gradle to make it work.

I saw that, I don't know how we can accomplish that.

fzurita:
I can probably take a look today sometime.

bladeoner:
That would be great, I will merge the commits to get a clean source.

fzurita:
I'm working on it:

https://github.com/mupen64plus-ae/mupen64plus-ae/tree/angrylion-rdp-plus-cmake

I will probably make a pull request upstream with CMakeLists modifications once it's all working right.

Right now the above branch runs, but it's very slow.

fzurita:
I went ahead and did it without CMake, I'll figure out why CMake is causing such a big slow down later. Done here: 729d9905e780414d115a78a379134ea7230ed7ff

fzurita:
Ok, so the slowness is because CMake is overriding my flags and actually building debug builds instead  of release builds. Debug builds lack optimizations, so it may be a none issue. I'll merge the CMake stuff if I see that it runs fasts if I make a release build.

bladeoner:
@fzurita, it looks like you have done it, great! :)

This makes it easier in the future to get it updated.

PR #787, opened 2020-01-25
UI improvements

alzurita:
UI improvements, it still needs to be tested in a TV interface


fzurita:
Thanks, this is a definite improvement in the looks of the app. I tested the TV interface and unfortunately, it's difficult to clock the "+" button to search for new ROMs, one has to scroll to the very bottom of the list to refresh the ROM list. Also, when putting focus on the search text field with a controller, there is no way to bring the on screen keyboard app.

alzurita:
In addition to the (+) fix, we'll need to look at the search bar text randomly disappearing. Maybe solving that will also solve the weird keyboard issue. :/

fzurita:
This has been merged. I'll fix any ideas with it.

PR #866, opened 2020-10-24
Add various 8BitDo profiles and adjust profile list

bladeoner:
- Add N64 RetroBit/8BitDo controller profile
- Add 8BitDo SN30 Pro/SN30 Pro+ controller profile
- Rearrange controller profile list


fzurita:
@bladeoner How did you verify these profiles? I tried the NVidia shield profile and the Z button and potentially the L are not working correct. The default "Android Gamepad" profile seems to be ok for the shield TV.

fzurita:
Ok, the L is mapped correctly, but the Z is mapped to R2 instead of L2. I guess I'm not a huge fan of that. Is the mapping for the shield controller similar to the other new ones you created?

bladeoner:
@fzurita, the other ones are mapped the same yes.

And what if we map Z to both L2 and R2?

bladeoner:
@fzurita I just assigned Z to both L2 and R2, when the build finishes please test with the Nvidia Shield controller if that gives the effect we want.

I also changed the same for the 8BitDo SN30 Pro/SN30 Pro+ controller profile

fzurita:
I would rather the profiles be kept consistent. Here is what I would do.

* Make R2 be Z in the default Android Gamepad profile (in addition to the existing L2)
* Remove the Nvidia Shield gamepad profile, since they would be mapped the same, it becomes redundant. 
* If the other controllers also have L2, map the Z to that as well, in addition to R2.

If you can do that, I will take the pull request.

bladeoner:
That's fine, I just adjusted this PR so that the Nvidia Shield profile maps Z to L2 and R2 and the 8BitDo SN30 Pro/SN30, so that leaves:
- Removing the Nvidia Shield profile
- Assign the Z button to both L2 and R2 in the Android profile.

I will look into that tomorrow when I changed the PR I will inform you, thanks in advance.

bladeoner:
@fzurita I adjusted the following:
- Assigned the Z button to both L2 and R2 in the Android profile.
- Removed the Nvidia Shield profile

Both L2 and R2 were already assigned to the new SN30 Pro/SN30 Pro+ profile and the N64 Retrobit profile has the regular N64 controller layout buttons assigned. There is a build running at this moment, if you could verify with the Android profile, I also verified it looks good.

fzurita:
I merged it, although I made some minor adjustments.

PR #880, opened 2020-12-29
Correct Set Environment

bladeoner:


fzurita:
What was wrong with this before?

bladeoner:
Setting ANDROID_NDK_PATH before with export was working in the Github Workflow but they changed it, I was running into this with another project. So I wanted to adjust that, but I'm not sure why it failed now.

fzurita:
```com.android.ide.common.signing.KeytoolException: Failed to read key AndroidDebugKey from store "/root/.android/debug.keystore": null```

fzurita:
It also seems to have downloaded NDK 22 instead of 21.

bladeoner:
I think it's better to keep it like it is for now, because it is working.

I was also lookin into using NDK 22, but I didn't managed to get it working yet they changed linking files:
https://github.com/android/ndk/wiki/Changelog-r22

I will close this PR.

PR #886, opened 2021-01-05
Only upload apk artifact

bladeoner:


fzurita:
Hmm, I still see a zip file being uploaded. How is it different?

bladeoner:
It's a minor adjustment, previously the complete folder was uploaded also the JSON file, now only the APK file.

fzurita:
Ok

PR #970, opened 2021-11-05
Revert change ec74d52

bladeoner:
Revert the following change because it causes slow downs: mupen64plus-ae@ec74d52


fzurita:
In what instances are you seeing slowdowns?

fzurita:
I reverted it through rebase.

bladeoner:
@fzurita I was testing Goldeneye and I saw extreme slow downs when the game starts and while playing after this change.

fzurita:
Interesting, and the issue went away after you reverted this specific commit?

bladeoner:
Yes, I'm also not sure what caused it in NDK 23b, but I saw that the b version of NDK 23 already has some changes more in line with NDK 24 then NDK 23a.

fzurita:
It may have to do with this: 

> Issue 1536: Make optimization flags used with CMake more consistent. Historically thumb release builds used -Oz, but AGP switched to using RelWithDebInfo for release builds in the latest release which was not using -Oz. To reduce per-arch differences and behavior differences compared to CMake's defaults, -Oz use was removed. You may see code size increases for armeabi-v7a due to this, but also increased optimization. To restore the prior behavior, add -Oz to your cflags.

bladeoner:
That can be I saw that with NDK 23b the file size was a little bit larger.

fzurita:
I added the -Oz option after updating to the latest NDK. Can you try that to see if you still get the slowdowns?

bladeoner:
@fzurita I saw you added the -Oz option but the build failed to compile.

fzurita:
Whoops, yeah, I forgot to add the new ndk version to the workflow, I force pushed an update.

fzurita:
And there are also build issues with the Oz flag... I went ahead and addressed those.

bladeoner:
@fzurita I tested the latest build, it is with slowdowns.

fzurita:
Which plugins are you using?

bladeoner:
I'm using the GlideN64-very-accurate profile and the default settings. The games I tested are Goldeneye 64 (U) and Mario Kart 64 (U) and the slowdowns are immediately after you start the game also in the menu's. I'm using a Nvidia Shield TV (2017).

When I use the build before the NDK bump I don't experience this:
https://github.com/mupen64plus-ae/mupen64plus-ae/actions/runs/1504008789

fzurita:
Is it low FPS or stutters?

bladeoner:
It stutters.

bladeoner:
It seems to be low on FPS for example with the stable one it's 59-60 fps in the menu's and with the latest build 35 fps.

bladeoner:
@fzurita do you experience any slowdowns?

fzurita:
I have not gotten around to see what is wrong yet. I have a Nvidia shield TV, I will probably try it this weekend.

fzurita:
I definitely see the slow downs... I'm trying to find the source now.

fzurita:
It seems gradle is ignoring -DCMAKE_BUILD_TYPE=Release when doing debug builds. I'll see if I can force the github action to do a release build.

fzurita:
Can you try the result of this github action?

https://github.com/mupen64plus-ae/mupen64plus-ae/actions/runs/1537646933

Unfortunately, the application ID changed, so it will install as a separate application from the debug version.

fzurita:
Actually, even in the release build, CMake flags are not being passed from grade to CMake... so you will probably still have the problem.

fzurita:
Ok, it looks like "-O3" seems to not be getting applied automatically anymore when CMAKE_BUILD_TYPE is Release. I went ahead and added all to all the CMake projects. Please try this build when you get a chance:

https://github.com/mupen64plus-ae/mupen64plus-ae/actions/runs/1537878795

bladeoner:
@fzurita I tested the latest and it seems to be running even better then before, thanks for fixing this.

With Goldeneye the first level I always heard some audio stuttering that seems to be gone now.

fzurita:
Great!

PR #971, opened 2021-11-05
Add Nintendo 64 Switch controller profile

bladeoner:
This will add a controller profile called "Nintendo 64 (Switch)" to the profile selection screen and adds the new official Nintendo 64 Switch Online controller.

The controller is not a generic bluetooth controller so I had to change the sensitivity to a higher value to get it working.

Just finished testing with multiple games, it works really good.


bladeoner:
@fzurita I tried using the following setting in the profile but that didn't seem to work:
sensitivity=156

The value above seems to be the value to get let the controller work properly but I had to do it like this:
sensitivity_x=156
sensitivity_y=156

fzurita:
Yeah, sensitivity doesn't exist anymore. It has to be specified separately for each axis.

PR #1000, opened 2022-02-03
Restructure files

bladeoner:
@fzurita I was just thinking about restructuring the files to make it more generic, I'm not sure if you want that.

I have a succesful build and just installed the emulator:
https://github.com/bladeoner/mupen64plus-ae/actions/runs/1789341752


fzurita:
Hmm, I'm not sure I'm ready for something like this.

Issue #997, opened 2022-01-31
File size increased a lot


bladeoner:
@fzurita I'm not sure why but the filesize increased from around 30 MB to 84 MB is that correct?

bladeoner:
It seems to have something to do after this commit:
[Update to the latest Android Studio version](https://github.com/mupen64plus-ae/mupen64plus-ae/commit/5dc63d4249a53e861a65e6ad7236fcb5d29d04f5)

fzurita:
Thanks for letting me know. I'll look into it.

fzurita:
This seems to be a bug with the latest Android Studio gradle version. It seems like even in a release build debug versions of libraries are being included no matter what I do...

bladeoner:
@fzurita it seems that with the last [commit](https://github.com/mupen64plus-ae/mupen64plus-ae/commit/79af453cf8cd7c942c9bd15841d8c13511b4125c) you fixed the size it is now around 30 MB, thanks!

PR #1001, opened 2022-02-04
Update Gradle

bladeoner:
This doesn't resolve issue #997, just an update.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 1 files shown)


bladeoner:
I need to rebase it before it can be added

bladeoner:
Rebase has been done.

fzurita:
I usually stick to the version that Android studio recommends since I know it has been tested well with it. Does going with this later version bring any advantages? If not, I would rather wait until an Android studio update recommends that specific version.

bladeoner:
It doesn't bring any advantages directly for us, but security wise it is patched for the log4j security bug that was what interested me to keep things safe.

fzurita:
I see, I would rather wait since as far as I know the log4j security bug doesn't directly affect the app. Also, you would think that Android studio would update to the later version on their own if that security vulnerability affected Android studio.

bladeoner:
Yes, that's true I will close this PR.

PR #1052, opened 2022-08-11
In game settings

OhThomas:
I added the ability to change settings in game through a settings menu accessible on the game sidebar drawer. This is very useful for testing settings out or switching things on the fly which is great for netplay so you can switch controllers, orientation, shaders, etc., without disconnecting. The changes I made are the following:

- Created a settings menu that can be used in game which allows the user to change preferences and see them update without having to close the game and go to the respective menu to change the setting and then reload the game before seeing the change take place

- Made changes to preferences available (combined 'Hide buttons when not used' with 'Number of seconds before hiding touchscreen buttons' to make it more coherent for the user and have the settings space less cluttered, made 'Animate joystick' have range of input from user so they can have it more or less animated to their choosing [it looks really good when updated frequently], added 'Game settings opacity' to let the user choose the opacity of the new in game settings menu, made the device vibrate when enabling 'Haptic feedback', and successfully importing a touchscreen graphic will change 'Button style' to Custom)

- Fixed many spelling errors


fzurita:
Wow, this look like a lot of work. I do see a lot of changes here though that probably don't belong here. For example, for the core, you would have to make a separate pull request at https://github.com/mupen64plus-ae/mupen64plus-core. I would then have to bring that change in here through a subrepo clone operation. I'll take a look at your changes.

app/src/main/java/paulscode/android/mupen64plusae/GalleryActivity.java
```diff
@@ -575,6 +646,22 @@ public void onPause() {
...
 
+        // We don't resume resetting the game if something happens here because we
+        // set the gallery grid to a blank state which could be saved if not loaded
+        // properly. So to avoid the issue we just stop the reset
+        if(mGameResetting){
+            // Could retain and show notification onResume but the data might not be retained
+            // so we'd have to write to disk which becomes a problem because the user could
+            // exit for good, we wouldn't know which could cause some confusing/false notifications
+            if(isChangingConfigurations())
+                runOnUiThread(() -> Notifier.showToast(this, R.string.toast_unexpectedChange));
```
fzurita, line 657: onPause should already be running on the UI thread, so you shouldn't have to call runOnUiThread here.

app/src/main/java/paulscode/android/mupen64plusae/jni/CoreTypes.java
```diff
@@ -69,6 +69,10 @@ enum m64p_core_param {
         M64CORE_EMU_STATE,
         M64CORE_VIDEO_MODE,
         M64CORE_SAVESTATE_SLOT,
+        M64CORE_AUDIO_INIT,
```
fzurita, line 72: New enums should probably be added at the end of the list to stay somewhat compatible with the upstream core.

mupen64plus-core/upstream/src/device/rcp/vi/vi_controller.c
```diff
@@ -170,6 +175,27 @@ void write_vi_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask
     masked_write(&vi->regs[reg], value, mask);
 }
 
+
+void resolution_reset_check(){
+    if(l_resolutionReset != 0){
+        // Making sure to pause once the core has started up;
+        // sometimes the graphics plugin finishes first
+        if(g_EmulatorRunning && !g_rom_pause) {
+            if(strncmp(get_gfx_name(),"GLideN64",7) == 0 && l_resolutionResetCoreCounter < 8) {
```
fzurita, line 184: In here you are adding to the core explicit knowledge of what the currently loaded video plugin is. This breaks the part of the core design which says that the core should have no knowledge of the graphics plugin. I would try to find another way of doing this.

mupen64plus-core/upstream/src/main/main.c
```diff
@@ -557,8 +557,21 @@ void main_state_inc_slot(void)
...
+    if (netplay_is_init() || l_FileName == NULL)
+        return;
+    savestates_set_job(savestates_job_load, savestates_type_unknown, l_FileName);
+}
+
 void main_state_load(const char *filename)
 {
+
+    if(filename != NULL) {
+        l_FileName = malloc(strlen(filename) + 1);
```
fzurita, line 571: I see a malloc here, but where is the free?

mupen64plus-core/upstream/src/main/main.c
```diff
@@ -1804,6 +1857,25 @@ m64p_error main_run(void)
     StateChanged(M64CORE_EMU_STATE, M64EMU_RUNNING);
 
     poweron_device(&g_dev);
+
+    DebugMessage(M64MSG_STATUS,"Rom header name = %s",ROM_PARAMS.headername);
+    // PS2 & PS1 needs additional frame count when resetting from in game settings (can use this
+    // count as default if more games get black screen when restarting from in game settings)
+    if( get_r4300_emumode(&g_dev.r4300) >= 2 && strncmp(ROM_PARAMS.headername,"POKEMON STADIUM",15) == 0 )
```
fzurita, line 1864: This code probably belongs in the front end. Also, if possible, I would try to find a generic solution for all games instead of game this which has a lot of changes of breaking or not working for games that were not tested.

mupen64plus-core/upstream/src/plugin/plugin.c
```diff
@@ -69,6 +69,8 @@ static const gfx_plugin_functions dummy_gfx = {
     dummyvideo_ReadScreen2,
     dummyvideo_SetRenderingCallback,
     dummyvideo_ResizeVideoOutput,
+    dummyvideo_PluginResolutionReset,
```
fzurita, line 72: Hmm, could this be done with ResizeVideoOutput instead of expanding the API?

fzurita:
You seem to have changed the plugin API, this is something I generally try to stay away from. It would be good if you can find ways of doing this without changing the API.

mupen64plus-core/upstream/src/device/rcp/vi/vi_controller.c
OhThomas: This was a temporary fix I seem to have forgotten to check which came from commit 'afa79e3c5b4f22ef9d8e939bf2e1290f9bec88b9', I'm testing without the get_gfx_name() function existing at all anymore and it still seems to work as intended, I'll continue to test when I settle some other things

mupen64plus-core/upstream/src/main/main.c
OhThomas: The free is on line 1926 when all the plugins are closed from the user exiting, do you think it could leak out some other ways?

mupen64plus-core/upstream/src/plugin/plugin.c
OhThomas: I don't believe so, I tried using that function when I was first trying to get the resolution setting to work and I can't remember what the problem was; but the resolution reset function pauses the game after it gets a readable frame from the video plugin when it starts which I applied to settings I couldn't get to work without restarting the activity, I'll look into some other solutions that maybe don't involve the video plugins

OhThomas:
I'm currently trying to get this to just work on the frontend or some other ways without changing the API, I changed the API to basically retrieve data for when things would initialize. I'm trying to find some less abruptive ways to get that data or some other solutions in general

fzurita:
If currently ```resizeWindow``` doesn't do what you need, it's possible you may adapt it to do so without changing the API.

OhThomas:
Hey I've made some changes and was wondering if they addressed these issues without raising more conflicts. I've changed how the feature interacts with the API, I've made the M64CORE_AUDIO_VOLUME and M64CORE_EMU_STATE return a negative value when not initialized to get rid of the M64CORE_AUDIO_INIT, M64CORE_EMU_MODE and M64CORE_EMU_MODE_INIT parameters. I've also made ResizeVideoOutput utilize setting the resolution; the video plugins still have a local counter that keeps track of how many frames have passed so the resolution and other changes can be seen immediately. I deleted a bunch of the expanding functions without taking away from the feature and wanted to hear your thoughts on these changes.

mupen64plus-core/upstream/src/api/frontend.c
```diff
@@ -69,6 +69,8 @@ EXPORT m64p_error CALL CoreStartup(int APIVersion, const char *ConfigPath, const
     SetDebugCallback(DebugCallback, Context);
     SetStateCallback(StateCallback, Context2);
 
+    g_EmuModeInitiated = 0;
```
fzurita, line 72: Make a pull request here for this change:
https://github.com/mupen64plus-ae/mupen64plus-core

fzurita: Well, do it for all core changes.

mupen64plus-core/upstream/src/device/rcp/vi/vi_controller.c
```diff
@@ -170,6 +170,24 @@ void write_vi_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask
...
 }
 
+
+void resolution_reset_check(){
+    if(g_ResolutionReset != 0){
+        // Making sure to pause once the core has started up;
+        // sometimes the graphics plugin finishes first
+        if(g_EmulatorRunning && !g_rom_pause) {
+            if(g_ResolutionResetCoreCounter < 12) {
+//                l_resolutionCounter++;
```
fzurita, line 180: This doesn't seem to be doing anything.

mupen64plus-video-angrylion-plus/upstream/src/output/screen.h
```diff
@@ -9,4 +9,5 @@ void screen_init(struct n64video_config* config);
 void screen_adjust(int32_t width_out, int32_t height_out, int32_t* width, int32_t* height, int32_t* x, int32_t* y);
 void screen_update(void);
 void screen_toggle_fullscreen(void);
+void screen_resolution_reset(void);
```
fzurita, line 12: This is going to be automatically removed next time I pull the latest angrylion-rdp-plus. I would recommend making a pull request here with proper justification:

https://github.com/ata4/angrylion-rdp-plus

mupen64plus-video-glide64mk2/upstream/src/Glide64/Main.cpp
```diff
@@ -1504,6 +1504,8 @@ EXPORT void CALL ReadScreen2(void *dest, int *width, int *height, int front)
   }
 }
 
+int    resolutionResetGlide = 0;
```
fzurita, line 1507: Make a pull request here:

https://github.com/mupen64plus-ae/mupen64plus-video-glide64mk2

mupen64plus-video-gliden64/upstream/src/mupenplus/MupenPlusAPIImpl.cpp
```diff
@@ -162,6 +162,8 @@ void PluginAPI::SetRenderingCallback(void (*callback)(int))
```
fzurita, line 162: For GLideN64, make a pull request here:

https://github.com/gonetz/GLideN64

mupen64plus-video-parallel/upstream/gfx_m64p.c
```diff
@@ -309,6 +309,10 @@ EXPORT void CALL SetRenderingCallback(void (*callback)(int))
```
fzurita, line 309: For this plugin, make the pull request here:

https://github.com/mupen64plus-ae/parallel-rdp-standalone

mupen64plus-video-rice/upstream/src/Video.cpp
```diff
@@ -51,6 +51,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 static void (*l_DebugCallback)(void *, int, const char *) = NULL;
 static void *l_DebugCallContext = NULL;
```
fzurita, line 52: For this, I will probably have to fork this repo to this project:

https://github.com/mupen64plus/mupen64plus-video-rice

fzurita:
I think you are heading in the right direction. I like that there are no API changes now. The main obstacle is that you made a lot of changes to plugins that are "subrepo clones" of other repos. So the next time I do a subrepo clone, the changes will be overwritten. 

I added links to the other repos where you need to make pull requests, they are not all owned by me, so you will have to provide proper justification for the changes and not break other platforms.

Also, this whole functionality needs to be disabled in the case where netplay is being used. Changing certain settings can cause a server and client to fall out of sync.

OhThomas:
Hey I've removed a lot of the changes in dependencies without removing any functionality; I still have a change in the gln plugin that's necessary for updating the resolution and was wondering if that came from another repository that would be better for the pull request to go to. I've also tested changing the settings I have available during netplay and find they don't cause desynchronization issues, I've made controller setting changes dependent on the netplay controller port assigned to the user to avoid sync issues. I can definitely disable changing settings during netplay but would advise keeping it for the benefits it provides, at least being able to change the controller input specifically.

PR #1054, opened 2022-08-18
Avoid buffer overflow read on 'm64pButtons'

### LINKED ISSUE #1037: [Issue]: Potential heap-buffer-overflow in plugin.cpp
### Describe the bug

I was able to detect a heap-buffer-overlfow error from Java_paulscode_android_mupen64plusae_jni_NativeInput_setState function in plugin.cpp file. I am attaching the AddressSanitizer [output here.](https://github.com/mupen64plus-ae/mupen64plus-ae/files/8921876/NativeInput_setState_crash_output.txt)

I believe the issue here is, there is no bound check in this function when accessing "androidButtonState" buffer using the "controllerNum" index variable.

### Steps to reproduce

Used libFuzzer based fuzzing to trigger and detect potential buffer overflow issue(s) in Java_paulscode_android_mupen64plusae_jni_NativeInput_setState function.

### Expected behavior

I would expect the application not to crash and the buffer access are handled properly.

### Relevant log output

_No response_

### Add screenshots

_No response_

Kishanthan:
This PR will fix #1037. This will avoid possible buffer overflow read on "m64pButtons" array.

### DIFF
--- a/mupen64plus-input-android/src/plugin.cpp
+++ b/mupen64plus-input-android/src/plugin.cpp
@@ -196,6 +196,11 @@ extern "C" JNIEXPORT void Java_paulscode_android_mupen64plusae_jni_NativeInput_s
         jdouble mp64pXAxis, jdouble mp64pYAxis, jboolean isDigital)
 {
     jboolean* elements = env->GetBooleanArrayElements(mp64pButtons, nullptr);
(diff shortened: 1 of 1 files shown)


fzurita:
This seems fine.

PR #1079, opened 2022-12-02
Add Pre-Release

bladeoner:
Hi @fzurita I've added Pre-release to the releases section, this will be updated and overwritten everytime a new PR to master is performed. I've added this because the Nightly builds will expire after some time:
image

The Pre-release will look like the picture below:
image


fzurita:
Great, thanks!

Issue #1017, opened 2022-03-13
[Issue]: Data loss while using the app (the fragment disappears after double screen rotation)


davirec:
### Describe the bug

When I perform these sequences of events, the fragment disappears after double screen rotation:

### Steps to reproduce

1. Open the app
2. Tap in the hamburger menu
3. Tap in Language
4. A screen similar to this will appear:

Before rotation:

<img src="https://user-images.githubusercontent.com/3637419/158076300-3e8957d0-e3f1-43da-9c25-a9de1534ee5e.png" width=30% height=30%>

5. Force the pause from the activity and then return to the activity (double screen rotation). Note that the fragment disappears.

### Expected behavior

The expected behavior is the return to the state of the previous image.

However, the fragment disappears.

After rotation:

<img src="https://user-images.githubusercontent.com/3637419/158076315-43d3cb8a-9014-4441-9bfa-03aaecefd019.png" width=30% height=30%>

This bug occurs in other screen elements: app version, faq, hardware info.

Note: 
- The pause from activity(1) can be performed in several ways. We suggest using a double screen rotation (starts in portrait, goes to landscape, and back to portrait) with this function previously enabled in the smartphone.
- Double screen rotation: It is executed twice to reach a state that should be exactly the same as the screen before if no data loss occurred.
- A data loss problem occurs when data is accidentally deleted or state variables are accidentally assigned with default or initial values.

Device:
- Device: Moto G30 with OS: Android 11
- Version: 3.0.246 (beta)
- Device language: English

1. https://developer.android.com/guide/components/activities/activity-lifecycle

fzurita:
Yep, I'm aware of this. A while back ago I went and fixed a bunch of dialogs that were critical but not all of them. Since they were not critical enough to be hurt by the data loss as you state.

There are still a few that need to be fixed.

davirec:
Thanks

davirec:
Hi fzurita,
in this activity, we found other bugs:

- GalleryActivity: lost the app version fragment, lost the faq fragment, lost the information fragment, restarted the activity and selected the first element, changed a view property.

<div id="ladoalado" style="display:flex;justify-content:center;">

<figure id="container">Before and after double screen rotation<br>
<img src="" width=20% height=20% >
</figure>

<figure id="container">
<img src="" width=20% height=20%>
</figure>

</div> 


<div id="ladoalado" style="display:flex;justify-content:center;">

[...]

PR #1082, opened 2022-12-17
Popups extending DialogFragment

OhThomas:
Addressing issue #1017 by extending Popups with DialogFragment and making LocaleDialog class to retain the fragments. I'm unsure what the problem is with the last two examples in the issue (ManageControllerProfilesActivity & ManageEmulationProfilesActivity) so I haven't addressed them with this pull request


fzurita:
Cool, this a good and simple change.

PR #1083, opened 2022-12-20
Multiplayer

Reganbunt:
I have a request can we have online/local multiplayer for egzample mario 64 kart


Reganbunt:
Also I think we need a cbox version of the app (just do it)

Reganbunt:
> Also I think we need a cbox version of the app (just do it)

Xbox not cbox

PR #1087, opened 2023-01-06
Optimize Github workflow and update Cmake to version 3.22.1

bladeoner:
@fzurita the best wishes for the new year, I've optimized the workflow and updated Cmake to version 3.22.1.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,6 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-    container: cirrusci/android-sdk:30-ndk
 
     steps:
     - name: Checkout mupen64plus-ae repo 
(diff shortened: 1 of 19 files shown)


bladeoner:
@fzurita when you have some time can you check this PR?

Thanks in advance.

fzurita:
Somehow this flew past my radar.

A lot of the changes you did were on upstream files, so next time I update an upstream library those files will get overwritten.

Ideally, you would make pull requests on upstream projects to get those updated and that way the changes would be retained next time I update a library.

In the end though, this didn't matter much because CMake is backwards compatible. It will work with CMakeLists files made for older versions of CMake, so I will still merge it.

Just keep in mind that some of the changes will be overwritten in the future if upstream projects don't update their CMake version.

PR #1139, opened 2023-06-10
Address undersized stick range and floating point error

kev4cards:
Currently, the control stick will not be able to produce values between (60, 60) or (61,61) and (68, 68), inclusive. This is due to the undersized scaling of the range when considering the maximum cardinal value as the maximum radius of the circle. Therefore, the diagonal (and the deadzone) also needs to be used in the initial scaling in order to ensure every value of the octagon is covered. Consequently, this value is now greater than the maximum cardinal value and requires a clamp after the octagon simulation to stop the excessive values.

After encountering issues with the diagonal not reaching its full value, the decision was made to add a small epsilon at the end just before casting to an integer to allow the diagonal to reach its full value. Same issue can happen for the deadzone. This method was chosen over rounding up because that would only reduce input and mean a small deadzone would be exited earlier than desired.


fzurita:
Merged to https://github.com/mupen64plus-ae/mupen64plus-ae/tree/improved_deadzone

PR #1159, opened 2023-11-13
Update node.js to version 16

bladeoner:
Hi @fzurita I hope your are doing well.

In the latest pipeline runs you see an warning about Node.js that is needs to be updated to a higher version.

Warning: _The following actions uses node12 which is deprecated and will be forced to run on node16: actions/setup-java@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/_

Example pipeline run: https://github.com/mupen64plus-ae/mupen64plus-ae/actions/runs/6845670591

To resolve the warning I have updated the Java action to v3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
         build-tools: 34.0.1
         platforms: android-34
 
-    - uses: actions/setup-java@v2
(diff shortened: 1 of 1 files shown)


fzurita:
Thanks!

PR #1201, opened 2025-02-03
Update CI actions

bladeoner:
Hi @fzurita,

I have updated the CI actions to update node.js to version 20 and fix the build error:
https://github.com/mupen64plus-ae/mupen64plus-ae/actions/runs/13090830122


fzurita:
Interesting, something appears broken. Unfortunately I don’t have time to debug right now. So it could be a bit.

Issue #1103, opened 2023-02-17
[Issue]: chromecast ROM path


kissklas:
### Describe the bug

Newer versions of android on the chromecast with google TV has locked off android/data while not having a native browser that works with apps that can access it. In other apps that use the android/media folder you can get around the issue by manually placing assets there, but for Mupen64 the only solution is to push roms to the data folder through developer mode.

A working way of manually writing in the path would be a work around. This is implemented in SNES9X but it doesnt seem to work, hence the need for manual placement into the default path in /media. 

### Steps to reproduce

Try browsing roms on a chromecast with android 12+, or transferring the roms to the default path. 

### Expected behavior

That an operating system comes with a file manager.

kissklas:
Oops, duplicate of  #1053. Sorry! hope this adds to the issue anyways.

TimeParadoxYT:
Is there a workaround for this? I've been looking for months, but I can't get it to work.

fzurita:
Unfortunately not unless you can root the device.

TimeParadoxYT:
That's sad. I'd rather not use retroarch because of the bad user experience. Thanks for the answer though.

stzokev:
SAF is not supported on Android TV and therefore not only file picker is missing, but file.uri, required by most emulators, is not supported. The workaround would be to have a fallback to manually enter the uri in a textbox/input field, instead of requiring a picker. That is an easy and graceful fallback!

WandLZhang:
I would like to echo OP @kissklas and @stzokev's messages, that is, **creating an option similar to SNES9X for manually specifying the file path**

Picture:

![Image](https://github.com/user-attachments/assets/f78c30f2-388e-4c54-bb9a-491a93426806)

The most critical functions would be ROM import and Game Data import. @fzurita if you are able to do this, much joy would be restored in my household that was robbed from us since 2022.

I am willing to fund this feature and donate, let alone leave a great Play Store rating. 

Part of me is curious to see what it would take to make a PR but I wouldn't know how this makes it into the Play Store build with lack of Android dev experience.

jschwartzenberg:
I installed the latest version from F-Droid and am running into this problem on a TV with Android 11. Other emulators and programs I installed through F-Droid can be given full access. There is a difference in the permissions that can be assigned. For instance with VLC I have this:
![Image](https://github.com/user-attachments/assets/fe1fab3c-9e74-4397-9f84-69f076e43ef4)

[...]

Issue #1053, opened 2022-08-11
[Feature request]: load game - app media folder


AngelofMe:
### Describe the feature

Android/media/org.mupen64plusae.v3.alpha

On Android 11+ add the app "media" directory to rootFileLocations() so that Android TV users without a working file picker have a public location to read/write files

stzokev:
Not sure why this is under enhancements, it is a solid bug or missing feature if you will. SAF is not supported on Android TV and therefore not only file picker is missing, but file.uri, required by most emulators, is not supported. The workaround would be to have a fallback to manually enter the uri in a textbox/input field, instead of requiring a picker. That is an easy and graceful fallback!

WandLZhang:
This should be addressed by PR #1242 (also linked from the related issue #1103).

The fix adds an "Enter Path" button to the Scan ROMs screen so you can manually type any folder path, plus `MANAGE_EXTERNAL_STORAGE` permission and enabling the legacy file browser on Android TV 11+ so it can browse the full filesystem.

Build-from-source instructions and usage details are in my [comment on #1103](https://github.com/mupen64plus-ae/mupen64plus-ae/issues/1103#issuecomment-4201461666).

WandLZhang:
Adding full build & sideload instructions here so folks don't have to click through:

### Build from source

```bash
# Clone the fork with the fix
git clone -b fix/google-tv-manual-path https://github.com/WandLZhang/mupen64plus-ae.git
cd mupen64plus-ae

# Make sure you have Android SDK with platform 36, build-tools 36.0.0, and NDK 26.1.10909125
# If using Android Studio, the SDK is usually at ~/Library/Android/sdk (Mac) or ~/Android/Sdk (Linux)
echo "sdk.dir=$HOME/Library/Android/sdk" > local.properties  # adjust path for your OS
./gradlew assembleDebug

# APK will be at app/build/outputs/apk/debug/Mupen64PlusAE-debug.apk
```

### Sideload to Chromecast / Google TV

- **Easiest:** Install [Send Files to TV](https://play.google.com/store/apps/details?id=com.yablio.sendfilestotv) on both your phone/computer and the TV. Send the APK to the TV's Downloads folder, then open it on the TV to install.
- **Via ADB:** `adb connect <YOUR_TV_IP>:5555 && adb install Mupen64PlusAE-debug.apk`

### How to use after installing

1. Open Mupen64Plus AE → tap the refresh/scan ROMs button
[...]

PR #1242, opened 2026-04-07
Add manual path entry for ROM loading on Android TV 11+

WandLZhang:
Hi @fzurita, this PR addresses issues #1103 and #1053 — the long-standing problem where Google TV / Chromecast with Google TV devices on Android 11+ can't browse to ROM folders because the SAF file picker is broken or unusable on those devices.

I took the approach you and others discussed in those issues — adding a manual path entry option similar to what SNES9X EX+ does.

**Changes:**

1. **"Enter Path" button** on the Scan ROMs screen — shows a text input dialog where users can type a folder path directly (e.g. `/storage/emulated/0/ROMs`). Uses the existing `Prompt.promptText()` utility and returns a `file://` URI, same as the legacy browser.

2. **`MANAGE_EXTERNAL_STORAGE` permission** — required on Android 11+ so that `java.io.File` access works outside the app sandbox. On first use of "Enter Path", the user is prompted to grant "All files access" in Settings.

3. **`useLegacyFileBrowser` now enables on all Android TV** (not just API < 30) — so the "Select Folder" button also uses the built-in legacy file browser instead of the broken SAF picker on these devices.

4. **Legacy file browser starts at storage root** when `MANAGE_EXTERNAL_STORAGE` is granted on API 30+, instead of being stuck in the app's sandboxed directory.

**Tested on:** Chromecast with Google TV (Android 12). Both "Enter Path" and "Select Folder" work — ROMs scan and appear in the gallery.

**Note on `MANAGE_EXTERNAL_STORAGE`:** This permission requires justification if publishing to the Play Store. It could be gated behind an Android TV check or a user-facing setting if needed. Happy to adjust the approach based on your preference.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -35,6 +35,8 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
+    <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE"
+        tools:ignore="ScopedStorage" />
     <uses-permission android:name="com.android.providers.tv.permission.READ_EPG_DATA" />
     <uses-permission android:name="com.android.providers.tv.permission.WRITE_EPG_DATA" />
     <uses-sdk tools:overrideLibrary="androidx.tvprovider" />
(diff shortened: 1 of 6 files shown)


fzurita:
My only concern is the MANAGE_EXTERNAL_STORAGE permission. If I were to publish an app update in the google play store, it would be a very hard uphill battle getting the app approved. It's already extremely difficult to get app updates approved and this would make it even harder.

fzurita:
I will go ahead and merge it and see how it goes.

